PR #25, opened 2019-10-01
Add missing multifile/1 directives

pmoura:
SWI-Prolog is actually quite permissive regarding missing `multifile/1` directives. But it's still good programming style to include them whenever defining clauses for multifile predicates.


GavinMendelGleason:
Merged to dev.  Cheers pmoura!

PR #44, opened 2019-11-16
WIP: Add first version of the test workflow

pmoura:
This pull request adds a test workflow that runs automatically on push and pull request events. It provides a solution for any contributors without a setup for running the tests (e.g. those like me on macOS) as pull requests will be automatically checked and rejected if tests fail.

Why the WIP tag?

1. The workflow is currently costly due to all the dependencies that are installed. It's possible that some of them are no longer required. You know best so please review the workflow file. If after removing any unnecessary dependencies the time to run the workflow is still to big (GitHub gives "only" 2000 free minutes to open-source projects), then I suggested to trigger the workflow only on pull request events.
2. The workflow succeeds even when the tests (run using `./start.pl test`) fail. The tests themselves need to be modified so that running them results in a non-zero exit value in case of failure.

Solving these two issues should allow this pull request to be merged.


rrooij:
Hey @pmoura, thanks a lot for your pull request. I do think however that we should just enable pull requests on our normal `.travis.yml` and disable the deploy stages for them. Because I have not implemented the conditions for the stages I disabled the pull request checking altogether. Do you think that is a better alternative than a seperate test for pull requests?

rrooij:
It should now already test pull requests. Can you submit a test pull request instance where you remove some trailing spaces from the README for instance?

pmoura:
I just merged the `dev` branch into this branch (`add_test_workflow`). Checks are running as expected.

rrooij:
Looks to be running fine. Would you mind if I close this one then?

pmoura:
Regarding using TravisCI (or other external to GitHub service) or GitHub own actions and workflows, there are pros and cons to both. GitHub actions and workflows provide a nice level of integration right from the repo (e.g. searchable logs). The test workflow itself can be rewritten easily to use Docker instead of the `setup-swi-prolog` action. I personally also currently use TravisCI for the `logtalk3` repo but I have started creating and  using GitHub actions and workflows in other repos. Both solutions work fine for preventing broken changes. Up to you. You can simply keep this PR open as an example of an alternative to TravisCI if the need arises.

pmoura:
P.S. If you're not familiar with GitHub actions and workflows execution, see e.g. this [demo](https://github.com/logtalk-actions/demo) repo. Go to the [commits list](https://github.com/logtalk-actions/demo/commits/master), click and hols on the green check mark, and click on the Details link.

PR #46, opened 2019-11-17
Fix test try/1 predicate to print failure message when test trows error

pmoura:


GavinMendelGleason:
Cheers! Much better now :D

PR #52, opened 2019-12-05
Remove redundant module explicit qualification from cors_catch/2 definition

pmoura:
As `cors_catch/2` is declared as a meta-predicate, the meta-argument is implicitly qualified by the runtime. Moreover, the predicate clause doesn't make use of the module qualifier.


rrooij:
Thanks!

PR #63, opened 2019-12-06
Fix woql_compile module missing use module directive

### LINKED ISSUE #59: Missing predicate: woql_context/1
The Logtalk make tool reports (on the `dev` branch):

```
*       woql_context/1
*         referenced from object woql_compile
*         in file /Users/pmoura/terminus-server/library/woql_compile.pl at line 458
```

The only definition I find for the predicate is in the `prefixes` module where the predicate is an exported dynamic predicate. Not clear **how** the predicate becomes visible in the `woql_compile` module as there's not reference there (notably, via a `use_module` directive) to the `prefixes` module.

pmoura:
Also fixes #59.


pmoura:
Sorry. Wrong branch.

PR #107, opened 2020-03-11
Make intersperse/2 predicate deterministic

pmoura:
Tested by comparing results with original version.

Note that this pull request doesn't change calls to this predicate. There are a few of them using `once/1` wrappers that should no longer be required.


GavinMendelGleason:
Cheers!

PR #108, opened 2020-03-11
Delete duplicated predicate export in terminus_bootstrap

pmoura:
Another Logtalk linter complaint 😛

### DIFF
--- a/library/terminus_bootstrap.pl
+++ b/library/terminus_bootstrap.pl
@@ -34,7 +34,6 @@
               repository_name_prop_uri/1,
               repository_remote_url_prop_uri/1,
               repo_type_document_prefix/2,
-              repo_type_document_prefix/2,
 
               repo_name_uri/4
           ]).


rrooij:
Is there by the way any guide on running the Logtalk linter? I'd love to have the linter running in our CI

pmoura:
@rrooij I currently use a hand-written `library/loader.lgt` file that loads the Prolog files in the `library` directory, compiling the modules as objects. I still need to patch a few files to avoid some errors (the Logtalk compiler is far more strict than SWI-Prolog but there are also some key differences between modules and objects). The order of the files loaded from the `loader.lgt` file attempts to minimize warnings and thus is not arbitrary. Happy to share my knowledge here (e.g. via an online call via Skype or similar; let me know). I can also commit the `loader.lgt` file (its presence in the repo should not cause any issues).

rrooij:
@pmoura Great to hear, would be nice to talk about this. Feel free to share your `loader.lgt` as well.

https://github.com/terminusdb/terminus-server/issues/51

pmoura:
@rrooij  I'm currently using a small `loader.lgt` file that allows applying the Logtalk linter:

```logtalk
% applies the Logtalk linter to all terminus-server
% modules files that are loaded 
%
% usage:
%	$ swilgt start.pl
%	?- {loader}.

:- initialization((
	set_logtalk_flag(report, warnings),
	logtalk_load([os(loader), types(loader)]),
	logtalk_load_context(directory, Directory),
	findall(
		File,
		(	current_module(Module),
			module_property(Module, file(File)),
			sub_atom(File, 0, _, _, Directory)
		),
		Files
	),
	forall(
		list::member(File, Files),
		(logtalk_compile(File, [unknown_entities(silent)]) -> true; true)
	) 
)).
```

You can experiment with it after installing Logtalk (latest version is recommended, currently at 3.36.0). Note that it will generate several warnings that are safe to ignore. It will also fail to lint a few modules due to compilation errors as they require manual patches to allow them to be compiled as objects. Those patches are not bugs in the modules but rather differences between objects and modules (notably, Logtalk is not based on a predicate prefixing mechanism as found in module systems). For example:

```text
!     Domain error: value prolog_frame_attribute(+,+,:) is not in domain meta_predicate_template
!       in clause for predicate report_curl_command/1
!       in file /Users/pmoura/terminus-server/core/util/test_utils.pl between lines 130-135
```

The patch here would be to add before the clause triggering the error:

```logtalk
:- meta_predicate(prolog_frame_attribute(*,*,*)).
```

But such patch would cause issues and thus no point in submitting it.

There are also a few false warnings regarding dict goals (no current workaround due to the way SWI-Prolog currently implements dicts). An example is:

```text
*     Goal is always false: A.instance_read_objects=[B]
*       while compiling object database
*       in file /Users/pmoura/terminus-server/core/transaction/database.pl between lines 238-272
```

When the goal is checked, the SWI-Prolog compiler have not yet compiled the dict call.

Anyway, give it a try. Happy to help clarify the warnings.

pmoura:
P.S. If you install the Logtalk git version (instead of the stable one), it includes some improvements to the compilation of modules as objects that improve the accuracy and thus usefulness of the linter.

PR #112, opened 2020-03-13
Fix missing argument in pcre:re_matchsub/4 predicate calls

pmoura:
There is no `pcre:re_matchsub/3` predicate.


rrooij:
Thanks a lot! Yeah, it is pretty annoying that there is no re_matchsub that doesn't take options.

PR #117, opened 2020-03-18
Add missing imports to resolve called predicates

pmoura:
The `global_prefix_expand2` predicate is called from `frame.pl` and the `merge_separator_split/3` is called from `database_utils.pl`.


rrooij:
Thanks for noticing. As far as I could see, we resolved this by importing them now.

PR #122, opened 2020-05-31
Fix typos

epost:


GavinMendelGleason:
Thanks for the contributions - the README was amended by hand.

PR #149, opened 2021-01-19
General copyediting: minor fixes, typos + Oxford commas

drkameleon:


rrooij:
Thanks!

luke-feeney:
great changes! Thou I would quibble about 'build from source' versus 'build from **the** source' - I know the latter is more grammatically correct, but the former sounds cooler...

PR #294, opened 2021-04-09
Fix duplicate results from queries.

GavinMendelGleason:
Found source of erroneous repeated solutions.


GavinMendelGleason:
Resolves #294

PR #300, opened 2021-04-13
Optimize race condition

GavinMendelGleason:
Fix the optimization race condition by using the version on the label to be updated in a test-and-set


rrooij:
Not to master plz!

PR #314, opened 2021-04-21
Adding better woql syntax error checking

GavinMendelGleason:
Currently woql syntax errors are not reporting with program navigable errors. This patch fixes the problem.


src/core/api/api_woql.pl
```diff
@@ -28,6 +28,8 @@
     context_extend_prefixes(Context,Woql_Prefixes,Final_Context),
 
     json_woql(Query, Final_Context.prefixes, AST),
-
-    run_context_ast_jsonld_response(Final_Context,AST,JSON).
+    
+    profile(
```
rrooij, line 32: This was probably done for debugging but should be left out now?

rrooij (approved):

PR #320, opened 2021-05-14
Capitalize "Git" in pdf

bl-ue:
https://git-scm.com/about/trademark


GavinMendelGleason:
This isn't for us thanks.

PR #325, opened 2021-05-17
allow TERMINUSDB_ADMIN_PASS from a file, especially for Docker's secrets feature

dwinston:
I lifted [docker-library's `file_env` shell function](https://github.com/docker-library/mongo/blob/b3edd5aacba7e89144cec6020ab34c13116ab44c/4.4/docker-entrypoint.sh#L34) to facilitate the common convention of allowing a `*_FILE` env var, i.e. `TERMINUSDB_ADMIN_PASS_FILE` in this case.

This makes it easier in my case to safely use version control for deployment configuration.

### DIFF
--- a/distribution/init_docker.sh
+++ b/distribution/init_docker.sh
@@ -1,8 +1,27 @@
 #!/bin/sh
-TERMINUSDB_ADMIN_PASS=${TERMINUSDB_ADMIN_PASS:-root}
 TERMINUSDB_SERVER_PORT=${TERMINUSDB_SERVER_PORT:-6363}
 TERMINUSDB_AUTOLOGIN=${TERMINUSDB_AUTOLOGIN:-false}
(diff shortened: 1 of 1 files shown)


rrooij:
I'm totally in favor of this. Thanks a lot for your contribution. I will let someone else have a look just to be sure.

dwinston:
I encountered an issue when deploying this and basing off the existing docker image:
```
FROM terminusdb/terminusdb-server:v4.2.1
WORKDIR /app/terminusdb
COPY terminus_init_docker.sh /app/terminusdb/distribution/init_docker.sh
CMD ["/app/terminusdb/distribution/init_docker.sh"]
```
where `terminus_init_docker.sh` is the new file in this PR. My issue was that the `#!/bin/sh` at top invokes the script using `dash`, which has an issue ("Bad completion") with the `file_env` scriopting. When I change the top line to `#!/usr/bin/env bash`, everything works great. I'm hesitant to suggest this change though, as it seems more major.

rrooij:
Changing it to bash should be fine :+1:

PR #326, opened 2021-05-19
allow TERMINUSDB_ADMIN_PASS_FILE

dwinston:
I lifted [docker-library's `file_env` shell function](https://github.com/docker-library/mongo/blob/b3edd5aacba7e89144cec6020ab34c13116ab44c/4.4/docker-entrypoint.sh#L34) to facilitate the common convention of allowing a `*_FILE` env var, i.e. `TERMINUSDB_ADMIN_PASS_FILE` in this case.

This makes it easier in my case to safely use version control for deployment configuration.

Use bash rather than dash, to avoid "bad completion" error (change the top line to `#!/usr/bin/env bash`).

This supersedes #325, with PR against `dev` as per @rrooij's recommendation.

### DIFF
--- a/distribution/init_docker.sh
+++ b/distribution/init_docker.sh
@@ -1,8 +1,27 @@
-#!/bin/sh
-TERMINUSDB_ADMIN_PASS=${TERMINUSDB_ADMIN_PASS:-root}
+#!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


rrooij:
Thanks a lot!

PR #377, opened 2021-07-23
updated json documentation in class_frame.

AstroChelonian:
You really were there and held my hand while doing it, so is there a point of explaining it?


GavinMendelGleason:
```% PL-Unit: employee_documents 
ERROR: /app/terminusdb/src/core/document/json.pl:5248:
	test doc_frame: failed

. done
```

This test doesn't pass. Should be trivial to fix in the pr.

PR #409, opened 2021-08-12
Same instance ID in the same transaction bug fixed

AstroChelonian:
1/3 issues resolved. The process exits with an error when two instances with the same ID are added with the same transaction.

remaining tasks: 

Duplication at POST
Duplication at PUT

### DIFF
--- a/src/core/api/api_document.pl
+++ b/src/core/api/api_document.pl
@@ -122,6 +122,7 @@ api_insert_document_(schema, Transaction, Stream, Id) :-
 
     do_or_die(Id = (Document.get('@id')),
               error(document_has_no_id_somehow, _)).
+
 api_insert_document_(instance, Transaction, Stream, Id) :-
     json_read_dict_stream(Stream, JSON),
     (   is_list(JSON)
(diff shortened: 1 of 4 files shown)


spl:
@AstroChelonian Can you please remove the whitespace differences? Also, try to make the diff as small as possible. In the end, I think your diff will just look something like:

```diff
+                         do_or_die(is_set(Ids), error(same_ids_in_one_transaction(Id), _))),
```

You can squash all commits into one.

matko:
This is just one place where double ids have to be checked though. we also need to do this for the case of document replacing, both in the case of a full replace through a POST and for the case of a partial replace through the PUT method.

AstroChelonian:
> This is just one place where double ids have to be checked though. we also need to do this for the case of document replacing, both in the case of a full replace through a POST and for the case of a partial replace through the PUT method.

Thanks @matko I shall address it in ticket 403 :)

src/core/api/api_document.pl
```diff
@@ -177,7 +177,7 @@
 
     with_transaction(Context,
                      forall(
-                         (   json_read_dict_stream(Stream,ID),
+                         (json_read_dict_stream(Stream,ID),
```
matko, line 180: one more whitespace issue here

PR #428, opened 2021-08-19
Remove excess space in document generation

spl:
Remove the unnecessary spaces before and after newlines. `git` doesn't like spaces before newlines.


spl:
Use <https://github.com/terminusdb/terminusdb/pull/428/files?w=1> to see that there are no changes.

PR #440, opened 2021-08-20
Add base and schema prefix URI checks

### LINKED ISSUE #416: Invalid URI prefixes cause schemas to be dropped.
From a clean storage:

```sh
# Create a database using `b/` and `s/` for the prefixes. Don't use `http://` here.
./terminusdb db create t --label label --comment comment --prefixes '{"@base":"b/","@schema":"s/"}'

# Create a schema. 
echo '{"@id":"Parent","@type":"Class","@key":{"@type":"Random"},"name":"xsd:string"}' | \
  xh 'http://localhost:6363/api/document/admin/t?author=a&message=m&graph_type=schema'

# Observe that the above schema is not found in a query
xh 'http://localhost:6363/api/document/admin/t?graph_type=schema'
```

No errors are reported.

spl:
* In schema checker and `create_db`
* Make `uri_has_protocol` a bit more stringent
* Fixes #416

### DIFF
--- a/src/core/document/schema.pl
+++ b/src/core/document/schema.pl
@@ -31,6 +31,7 @@
 :- use_module(core(util)).
 :- use_module(core(transaction)).
 :- use_module(core(triple)).
+:- use_module(core(triple/literals)).
 :- use_module(core(query), [has_at/1, compress_dict_uri/3]).
 
 :- use_module(json). % This feels very circular.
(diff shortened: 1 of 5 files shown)


src/core/document/schema.pl
```diff
@@ -183,6 +180,23 @@
...
 
+refute_schema_context_system_prefix(Schema, Prefix_Name, Witness) :-
+    (   xrdf(Schema, 'terminusdb://context', Prefix_Name, Prefix_Value)
+    ->  (   literal_to_string(Prefix_Value, String),
+            uri_has_protocol(String)
+        ->  fail
+        ;   Witness = witness{
+                          '@type': context_system_prefix_is_not_a_uri,
+                          prefix_name: Prefix_Name,
+                          prefix_value: Prefix_Value
```
matko, line 191: Prefix_Value could be a non-jsonifiable thing. You should make sure that values that go into witnesses can be converted into json.

matko (changes requested):

PR #452, opened 2021-08-24
Ensure that we have proper default prefixes when creating db from cli

### LINKED ISSUE #448: Schema check failure not caught in DB create for API
This should return a pretty error. Also it would be nice if base and schema prefixes could be specified separately on the command line and with some sensible defaults.

```bash
gavin@titan:~/dev/terminusdb$ terminusdb db create test
  [15] throw(error(schema_check_failure(...),_18264))
  [13] db_create:create_db(query_context{all_witnesses:false,authorization:'terminusdb://system/data/admin',bindings:[],default_collection:system_descriptor{},files:[],filter:type_filter{types: ...},prefixes:_18380{'@base':"terminusdb://system/data/",'@schema':"http://terminusdb.com/schema/system#",'@type':'Context',api:'http://terminusdb.com/schema/api#',owl:'http://www.w3.org/2002/07/owl#',rdf:'http://www.w3.org/1999/02/22-rdf-syntax-ns#',rdfs:'http://www.w3.org/2000/01/rdf-schema#',sys:'http://terminusdb.com/schema/sys#',vio:'http://terminusdb.com/schema/vio#',woql:'http://terminusdb.com/schema/woql#',xdd:'http://terminusdb.com/schema/xdd#',xsd:"http://www.w3.org/2001/XMLSchema#"},selected:[],system:system_descriptor{},transaction_objects:[...],update_guard:_18358,write_graph:system_graph{type:instance}},'terminusdb://system/data/admin',admin,test,'','',true,false,_18458{doc:'terminusdb:///data/',scm:'terminusdb:///schema#'}) at /home/gavin/dev/terminusdb/src/core/api/db_create.pl:156
  [12] catch(cli:create_db(...,'terminusdb://system/data/admin',admin,test,'','',true,false,...),error(schema_check_failure(...),context(_18540,_18542)),cli:do_or_die(...,...)) at /home/gavin/.swivm/versions/v8.2.2/lib/swipl/boot/init.pl:530
  [11] catch_with_backtrace(cli:create_db(...,'terminusdb://system/data/admin',admin,test,'','',true,false,...),error(schema_check_failure(...),context(_18628,_18630)),cli:do_or_die(...,...)) at /home/gavin/.swivm/versions/v8.2.2/lib/swipl/boot/init.pl:580

Note: some frames are missing due to last-call optimization.
Re-run your program in debug mode (:- debug.) to get more detail.
Error: schema_check_failure([witness{'@type':context_missing_system_prefix,prefix_name:'http://terminusdb.com/schema/sys#base'}])```

### PR DESCRIPTION
**Pull request by matko**
Fixes #448 .

### DIFF
--- a/src/cli/main.pl
+++ b/src/cli/main.pl
@@ -897,7 +897,7 @@ run_command(db,create,[DB_Path],Opts) :-
     option(schema_prefix(Schema_Prefix), Opts),
     option(prefixes(Prefixes_Atom), Opts),
     atom_json_dict(Prefixes_Atom, Prefixes, []),
-    put_dict(Prefixes, _{doc : Data_Prefix, scm : Schema_Prefix}, Merged),
+    put_dict(Prefixes, _{'@base' : Data_Prefix, '@schema' : Schema_Prefix}, Merged),
     api_report_errors(
         create_db,
         create_db(System_DB, Auth, Organization, DB, Label, Comment, Schema, Public, Merged)),

### CODE REVIEW

**Comment by spl** [comment]
Does this change mean we always use default prefixes when creating a database?

**Comment by matko** [comment]
> Does this change mean we always use default prefixes when creating a database?

No. We use the default prefixes, unless they are overridden. You can override either by using the `--data-prefix` and/or `--schema-prefix`, or you can specify your prefix document like before, with an '@base' and/or an '@schema'. If any of that is present, it takes precedence over the defaults.

this should ensure that there's always a value for the base and schema prefix though.

**Comment by spl** [comment]
Before this change, I was unable to do `./terminusdb db create testdb` without passing `--prefixes`. After this change, I don't need `--prefixes`, and I see something when I query the schema:

```json
{
    "@base": "terminusdb:///data/",
    "@schema": "terminusdb:///schema#",
    "@type": "@context"
}
```

So, my understanding is that we're using default prefixes.

GavinMendelGleason:
That's correct.

matko:
> So, my understanding is that we're using default prefixes.

We are using them, but not always was my point. You can still override them but we have a good default now.

spl:
Right. I didn't mean always as in never allowing overriding. We just didn't have them at all before this change. 😬

Nonetheless, very happy to see this! I always wondered why I had to type so much to create a database!

PR #479, opened 2021-08-27
Fix docs workflow for new branch names

rrooij:
The docs workflow did not reflect the branch name changes from `master` to `main`. It also did not include tagged releases.

For now, the workflow creates a folder in docs based on its version. For instance:

- A push to main will save the docs to a folder called `main`
- A tagged release will make a folder containing its version

Am I missing something? I've been thinking that the latest tagged release should always be moved to the root as well.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,10 +2,12 @@ name: Docs
 on:
   push:
     branches:
-      - master
+      - main
       - rc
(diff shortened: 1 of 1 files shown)


spl:
We've moved our documentation to <https://github.com/terminusdb/terminusdb-docs>, so this isn't needed now.

PR #499, opened 2021-09-01
Id generation

GavinMendelGleason:
Add lexical subdocument addressing


matko:
Merging this pull request will invalidate all existing deployments, since the id generation is incompatible with the old way of doing things, and the system graph has changed too.

PR #519, opened 2021-09-05
Foreign Types

GavinMendelGleason:


src/core/document/schema.pl
```diff
@@ -599,6 +604,9 @@
 type_descriptor(_Validation_Object, Class, unit) :-
     is_unit(Class),
     !.
+type_descriptor(Validation_Object, Class, foreign(Class)) :-
```
matko, line 607: Should a foreign descriptor not also include the foreign type?

matko:
If this is to be used for domain sharding, we probably also want to enforce that any objects inserted in this data product do not have a type corresponding to any foreign type registered in this data product.

PR #523, opened 2021-09-06
Revert "Catch json(unexpected_end_of_file) and throw malformed_json_payload"

GavinMendelGleason:
Reverts terminusdb/terminusdb#522


GavinMendelGleason:
1) The double catch is discouraged
2) No test, so no way to know if this has suffered regression making it awkward to refactor

Issue #520, opened 2021-09-06
Underscores in names not encoding properly leading to collision dangers.


GavinMendelGleason:
Perhaps we can switch out the uri_encoded with the old www encoding strategy?

Cheukting:
Will test with Python client when fixed

GavinMendelGleason:
Fixed in #536

PR #536, opened 2021-09-07
Fix underscore inclusion in id fragments.

GavinMendelGleason:


GavinMendelGleason:
fixes #520

matko (approved):
Great work! Good catch on the enum too.

PR #539, opened 2021-09-08
fix foreign type test that was failing due to new id scheme

matko:
This fixes the unit test that was failing due to id changes.

However, there's something weird going on here. The id references in the document do not match the actual ids, as one has the + escaped and the other does not. This is probably unintended behavior so I'm not yet merging.

@GavinMendelGleason please have a look.


matko:
explanation is an intentional double encoding of a foreign id. nothing out of the ordinary.

PR #549, opened 2021-09-09
Idempotent insert

### LINKED ISSUE #544: proposal: new option ignore_duplicates on document insert
I would like a flag that allows document insertion to ignore duplicates, while still erroring on differences.

When data originates from another system, the workflow for getting that into terminusdb is by doing some export in that other system, followed by an import into terminusdb. For example, take a bank statement. This will contain all transactions for a particular time range. Banks typically allow you to export this information in various formats, such as CSV.

When such statements overlap, then on import, terminusdb will give an error. This can happen for example if you had previously imported an incomplete monthly statement, and then later on try to import the complete statement.

It'd be good if in this case, the document insertion endpoint could be told to verify that the old document for a particular id matches the new document, and simply ignore the new document if these match.

matko:
This is still a work in progress.

Currently this will still fail if subdocuments have random ids but are otherwise equivalent.

Resolves #544.

### DIFF
--- a/src/core/api.pl
+++ b/src/core/api.pl
@@ -115,7 +115,7 @@
               api_generate_documents_by_type/10,
               api_generate_documents_by_query/11,
               api_get_document/8,
-              api_insert_documents/9,
+              api_insert_documents/10,
               api_delete_documents/7,
               api_delete_document/7,
               api_replace_documents/8,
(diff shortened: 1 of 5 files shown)


GavinMendelGleason (commented):
`status = not_present` should presumably be `Status = not_present`

PR #553, opened 2021-09-09
when inserting and updating documents, return ids compressed

### LINKED ISSUE #548: ids for newly inserted documents are no longer being compressed
When inserting documents, you get a list of ids back. These used to be compressed, but this no longer seems to be the case.

Example:
```
$ curl -u "admin:root" -X POST -H "Content-Type: application/json" 'http://localhost:6363/api/document/admin/foo?graph_type=instance&author=x&message=x&ignore_duplicates=true' --data-binary '{"@type": "Thing", "x": "asdf"}'
["terminusdb:///data/Thing/asdf" ]
```

The expected output was
```
["Thing/asdf"]
```

matko:
Fixes #548.


matko:
Closing as the pull request has gone stale and the corresponding issue is now closed.

PR #559, opened 2021-09-11
check type kind on schema elaboration and error if it is unknown

### LINKED ISSUE #437: Inserting classes of unknown type into the schema graph is not checked properly
It is possible to insert a type definition into a schema like the following:
```
{"@id":"Moooo",
 "@type":"This_Type_Has_No_Meaning",
 "name":"xsd:string"}
```
Even though the given type is unknown in our schema language this will still be accepted. For the above example, it will insert a type triple, as well as a triple for the `name` property definition, so it appears this is treated like a Class. Subsequent document queries by id can retrieve this definition, but retrieving the full schema will skip over it, since this only retrieves classes of known types.

We should make sure that everything in the schema has a type known to the checker, so we don't end up with garbage in our schemas. This is especially important to catch typos like 'Claass' instead of 'Class'.

matko:
Resolves  #437.

### DIFF
--- a/src/core/api/api_error.pl
+++ b/src/core/api/api_error.pl
@@ -1206,6 +1206,24 @@ api_document_error_jsonld(Type, error(casting_error(Value, Destination_Type, Doc
                               'api:document' : Document },
              'api:message' : Msg
             }.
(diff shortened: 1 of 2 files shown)


matko:
Superseded by #564.

PR #564, opened 2021-09-12
Last chance hotel for schema elaboration

GavinMendelGleason:
This will at *minimum* alert the user that the schema document being added is not viable rather than silently dropping the insert.


matko:
Looks like we did some overlapping work in regards to recognizing unknown schema types (#559). Your solution for this is probably better though.

GavinMendelGleason:
Oh no! I should have looked more closely...

matko (approved):

PR #588, opened 2021-09-15
Document explaining our release process

matko:
This is both a description and a proposal for how to handle our release process.

Leaving this as a draft request so people can have a look.


spl:
Nice! Some initial comments:

1. Uppercase day names and “european”. (I know it's different in Dutch. :wink:)
2. Can we use a lowercase file name, `release.md`? I feel that all these uppercase file names are trying really hard to get my attention, but they don't need it at the moment.
3. I think you might be using minor when you mean patch in places. But it's not a big deal.
4. Use the prefix `v` in your versions.

PR #595, opened 2021-09-15
Properly handle key fields that are type containers

### LINKED ISSUE #570: Ensure Lexicals over optionals work
There's some uncertainty over what lexicals over optionals do right now, and if it is the right thing. Better get that checked.

matko:
Optionals, sets, arrays and lists were not working properly with key generation.

This now follows the following algorithm:
- For all 4, if there's just a single value, it is treated exactly as if this was a required field. This ensures that the same id is generated, so that in the future, it may be possible to switch between key strategies without having to convert existing objects that are already in line with the new format, such as when going from optional to set, or from required to optional.
- For all 4, if there is no value, the out-of-band value ~~'+'~~ '+none+' (unencoded) is used
- For the list-like types with more than one value, the values are concatenated with an unencoded '++' between them. In case of sets, the list is first sorted to ensure that documents who order the list differently but contain the same values still get the same id.

Work in progress as I explore some more edge cases.

Resolves #570.

### DIFF
--- a/src/core/document/json.pl
+++ b/src/core/document/json.pl
@@ -59,13 +59,20 @@
 encode_id_fragment(Elt, Encoded) :-
     ground(Elt),
     !,
-    format(atom(Fragment), '~w', [Elt]),
-    uri_encoded(segment, Fragment, Part_Encoded),
-    re_replace('\\+'/g, '%2B', Part_Encoded, Encoded).
+    (   Elt = optional(none)
+    ->  Encoded = "+"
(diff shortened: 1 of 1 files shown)


matko:
I replaced the oob value for missing property with `+none+` from its previous `+`. The reason is that otherwise, it is possible to construct a `++` out of multiple missing optional keys, which may conflict with the way lists key values are now represented. It probably wasn't an actual issue, but better safe than sorry.

Undrafting this pull request.

PR #606, opened 2021-09-23
Implement a new logging strategy

matko:
This pull request will replace the existing log with a new type.

The new logger allows structured logging of json structures by severity, and will make an effort to associate each log entry with an ongoing request.

Through the environment variable `TERMINUSDB_LOG_LEVEL`, the desired log level can be set. this can be one of DEBUG, INFO, NOTICE, WARNING or ERROR. The default is INFO. Anything below the desired log level is omitted from the log.

Through the environment variable `TERMINUSDB_LOG_FORMAT`, the desired log format can be set. By default this is `text`, which will just print the log severity, a timestamp and the log message (provided those fields exist in the structured log record). By setting it to `json`, the full structured log record is printed instead.

This log format should be compatible with google cloud logging.

Currently, log records are made for the following cases:
- request start (info)
- request end (info)
- descriptor open (debug)
- descriptor commit (debug)
- uncaught errors (error)

### DIFF
--- a/src/bootstrap.pl
+++ b/src/bootstrap.pl
@@ -18,7 +18,6 @@
 :- use_module(core(query)).
 :- use_module(core(api)).
 
-:- use_module(library(http/http_log)).
 :- use_module(library(qsave)).
 
 
--- a/src/config/terminus_config.pl
+++ b/src/config/terminus_config.pl
@@ -36,8 +36,11 @@ initialize_system_ssl_certs :-
 server_protocol(Value) :-
     Value = http.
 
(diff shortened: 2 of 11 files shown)


matko:
Open question is whether or not the http requests are logged correctly. Currently the log entries are logging some of the info on at the start of a request, and some of the info at the end of a request. It could be that log aggregators expect one big message with all the info in it, in which case we'll need to merge out the existing log data and have different (debug) records for request start and request finish.

GavinMendelGleason (approved):
Great stuff!

PR #607, opened 2021-09-23
Log improvements

matko:
This is a last handful of changes to make logging slightly better.


matko:
I'm still thinking we ought to merge request and response in one big log entry with all the information, to be logged at info level. The request start and request end can then be demoted to debug.

But this doesn't necessarily need to happen for this version.

PR #626, opened 2021-09-30
Further log improvements

### LINKED ISSUE #615: Log authentication information
When a user is authenticated, it'd be good if we logged who they are and how they authenticated, so we can associate operations in the log with actual people.

### LINKED ISSUE #616: Allow operation id to be submitted in a header for logging purposes
Currently in the json log format, we have an operation id which is meant to work with google cloud logging, associating all the log entries of a particular request together. However, it is very possible that terminusdb requests themselves are part of a larger cloud operation, such as in terminusx where multiple components together make up an operation. In this case, it'd be better if the operation id was not determined by terminusdb itself, but was submitted to it by the caller.

I propose we add a HTTP header, like `X-Operation-ID`, where an operation id can be submitted. The current operation-id which is really just a request id can move to its own field.

Other components using google's structured logging could then use the same operation id to group log entries accross different cloud bits.

@Francesca-Bit what do you think?

### LINKED ISSUE #625: Merge http request log into a single line
Currently, HTTP requests are logged as two lines - one line when the request comes in, and a second line after the response has been sent.
This is somewhat atypical. It seems like google logs expect HTTP logs to be only a single line containing both the request and response information.

To support this, we need to save our logging information for the request for after the response, then send both as one record. An additional advantage of doing this is that we can calculate the request latency and include that in the log.

matko:
This pull request will resolve a bunch of logging issues:

- [x] merge request and response log in one line (Fixes #625)
- [x] use operation id if supplied in a header (Fixes #616)
- [x] log authentication information (Fixes #615)

### DIFF
--- a/src/core/util.pl
+++ b/src/core/util.pl
@@ -213,7 +213,8 @@
               notice_log_enabled/0,
               info_log_enabled/0,
               debug_log_enabled/0,
-              generate_operation_id/2
+              generate_operation_id/2,
+              saved_request/5
           ]).
 
 % note: test_utils is intentionally omitted
--- a/src/core/util.pl
+++ b/src/core/util.pl
@@ -194,20 +194,26 @@
 
               % json_log.pl
               json_log_error/1,
+              json_log_error/2,
               json_log_error_formatted/2,
               json_log_warning/1,
+              json_log_warning/2,
               json_log_warning_formatted/2,
               json_log_notice/1,
+              json_log_notice/2,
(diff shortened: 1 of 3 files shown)


matko:
One annoyance about this change is that I'm forced to save some request info in a dynamic predicate, so that it is still available after the response has been sent. I just hope this won't lead to a noticable slowdown.

GavinMendelGleason (commented):
Should we ask @JanWielemaker about the ip?

PR #642, opened 2021-10-04
Rename Nightly to Non-Docker

spl:
* It's a less confusing name, since “nightly” could also mean the nightly release of Rust, for example.
* Run every night instead of every weeknight. It doesn't hurt.


rrooij:
I really like this. Nightly was a confusing name.

PR #645, opened 2021-10-05
Print swipl --version

spl:


spl:
I triggered a [run](https://github.com/terminusdb/terminusdb/actions/runs/1306684142) of the action. When that finishes successfully, I'll merge.

spl:
The `macos` job appears to be stalled on the tests, but both jobs printed the version of `swipl`.

PR #654, opened 2021-10-05
Add endpoint for front end.

GavinMendelGleason:
Adding an endpoint for recovery of organizations and associated databases.

### DIFF
--- a/src/core/api/api_user_organizations.pl
+++ b/src/core/api/api_user_organizations.pl
@@ -1,0 +1,60 @@
+:- module(api_user_organizations, [
+              user_organizations/3
+          ]).
+
+:- use_module(core(util)).
+:- use_module(core(account)).
+:- use_module(core(query)).
+:- use_module(core(document)).
+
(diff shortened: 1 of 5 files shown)


matko (dismissed):

src/core/api/api_user_organizations.pl
```diff
@@ -0,0 +1,60 @@
...
+              user_organizations/3
+          ]).
+
+:- use_module(core(util)).
+:- use_module(core(account)).
+:- use_module(core(query)).
+:- use_module(core(document)).
+
+user_organizations(System_DB, Auth, Result) :-
+    % TODO, need auth check.
```
matko, line 11: This TODO is actually not valid. The auth check is implicit in the query.

matko (approved):

spl:
I triggered the CI manually. Here is the [run](https://github.com/terminusdb/terminusdb/actions/runs/1308774802).

PR #688, opened 2021-10-15
Add integration test suite

spl:


matko:
Nice!
I'm not sure how the CI changes will affect things, maybe @rrooij can have a quick look?

PR #707, opened 2021-10-19
Speed up using thread local tabling

GavinMendelGleason:
Some experimental tabling for speedups


GavinMendelGleason:
Would probably be a better and may have significant impact if the predicates were refactored to look at the layer rather than the validation object or transaction object. This will allow the post-check to share work with the document transformation and could have impacts on performance.

PR #732, opened 2021-10-25
Document path

GavinMendelGleason:
Adding templating and other features needed to make an effective document path query as per discussion #729

### DIFF
--- a/src/core/query/json_woql.pl
+++ b/src/core/query/json_woql.pl
@@ -179,17 +179,19 @@ json_type_to_woql_ast('Distinct',JSON,WOQL,Path) :-
     WOQL = distinct(WOQL_Args,Sub_WOQL).
 json_type_to_woql_ast('And',JSON,WOQL,Path) :-
     _{and : Query_List} :< JSON,
-    maplist({Path}/[Q,W]>>(
-                json_to_woql_ast(Q,W,[and
+    index_list(Query_List, Indexes),
+    maplist({Path}/[Q,I,W]>>(
+                json_to_woql_ast(Q,W,[and,I
                                       |Path])
-            ), Query_List, WOQL_Queries),
+            ), Query_List, Indexes, WOQL_Queries),
     xfy_list(',',WOQL,WOQL_Queries).
 json_type_to_woql_ast('Or',JSON,WOQL,Path) :-
     _{or : Query_List} :< JSON,
-    maplist({Path}/[Q,W]>>(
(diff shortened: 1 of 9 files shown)


src/core/query/json_woql.pl
```diff
@@ -759,16 +760,15 @@
...
     variable_list(Spec,WSpec),
-    variable_list(Obj,Template),
-    % Add only one variable if only one in template
-    (   Template = [V]
-    ->  WObj = V
-    ;   WObj = Template),
+    % Backwards compat with "variable list"
+    (   variable_list(Template,WObj)
+    ->  true
+    ;   json_to_woql_ast(Template,WObj,[template|Path])),
```
GavinMendelGleason, line 771: This is too abstract - since we can't actually deal with an arbitrary query here. Instead we need a datatype to ast conversion predicate, and the cases of list, dictionary, variable, node, etc. need to be handled there.

src/core/query/json_woql.pl
```diff
@@ -814,15 +814,15 @@
...
 json_type_to_woql_ast('Dot',JSON,WOQL,Path) :-
-    _{dictionary : Dictionary,
-      dictionary_key : Key,
-      dictionary_value : Value
+    _{document : Dictionary,
+      field : Key,
+      value : Value
      } :< JSON,
-    json_to_woql_ast(Dictionary, WDictionary, [dictionary_key
+    json_to_woql_ast(Dictionary, WDictionary, [document
```
GavinMendelGleason, line 821: Use the dictionary conversion here... this can't be anything!

src/core/query/json_woql.pl
```diff
@@ -1531,4 +1547,28 @@
...
+                                  "field" : "employed",
+                                  "value" : { "@type" : "Value",
+                                              "data" : true }},
+                                { "@type" : "FieldValuePair",
+                                  "field" : "father",
+                                  "value" : { "@type" : "Value",
+                                              "variable" : "Father"}}]}}',
+    atom_json_dict(JSON_Atom, JSON, []),
+    json_woql(JSON,WOQL),
+    WOQL = json{'@type':"User",employed:true,father:v('Father'),name:"Jim"}.
```
GavinMendelGleason, line 1572: Should this not be a different label?

GavinMendelGleason (commented):
Let's fix these first...

matko (approved):

PR #750, opened 2021-10-30
Adding update or insert functionality

GavinMendelGleason:
Replace document now takes a flag "create" which if set to "true" will create a new document resource if it does not exist already.

### DIFF
--- a/src/core/document/json.pl
+++ b/src/core/document/json.pl
@@ -19,11 +19,13 @@
               insert_document/3,
               replace_document/2,
               replace_document/3,
+              replace_document/4,
               nuke_documents/1,
               insert_schema_document/2,
               delete_schema_document/2,
               replace_schema_document/2,
               replace_schema_document/3,
+              replace_schema_document/4,
               nuke_schema_documents/1,
               database_prefixes/2,
               insert_context_document/2,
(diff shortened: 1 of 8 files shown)


src/core/document/json.pl
```diff
@@ -2019,28 +2021,33 @@
...
 
-replace_document(Transaction, Document, Id) :-
+replace_document(DB, Document, Id) :-
+    replace_document(DB, Document, false, Id).
+
+replace_document(Transaction, Document, Create, Id) :-
     is_transaction(Transaction),
     !,
     json_elaborate(Transaction, Document, Elaborated),
     get_dict('@id', Elaborated, Id),
```
matko, line 2033: This requires the id to be present. If we want this to work like an insert for new objects, shouldn't we also have it generate the id when it is missing? Otherwise, the burden is on the client code or the user to generate the proper id that would otherwise have been generated during a normal insert call.

matko (commented):
I'm worried about the requirement that ids are present in the submitted documents. Let's discuss.

src/core/document/json.pl
GavinMendelGleason: This is after elaboration, so the id will exist at this point.

matko (approved):

Issue #347, opened 2021-06-28
Path regular expression is inconsistent to expectations for repeated "any predicate"


hoijnet:
**Describe the bug**
A path regular expression "(*)+" or "(*){1,2} does not identify/traverse more than one level from source to target which would be expected considering that "(property_name_level_1|property_name_level_2)+" does traverse more than one level.

**To Reproduce**
1. Go to TerminusDB Console Query tab
2. Enter any of the following:
   * Traverses two levels: ```WOQL.path("v:source", "(propertyname_firstlevel|propertyname_secondlevel){1,2}", "v:target", "v:traversed")```
   * Does not work, expecting same traversals or more than above: ```WOQL.path("v:source", "(*){1,2}", "v:target", "v:traversed")```
   * Traverses two levels: ```WOQL.path("v:source", "(propertyname_firstlevel|propertyname_secondlevel)+", "v:target", "v:traversed")```
   * Does not work, expecting same traversals or more than above: ```WOQL.path("v:source", "(*)+", "v:target", "v:traversed")```
3. Click Run Query

**Expected behavior**
For the ones that work I got the correct result, traversal from ```source-->intermediary-->target``` and the path in v:traversed.

For the ones that does not work, when using "any predicate, *", only one level is traversed: ```source-->intermediary```, and only the path for the source-->intermediary traversal is present in the result.

**Screenshots**
N/A

**Info (please complete the following information):**

![image](https://user-images.githubusercontent.com/144317/123678461-9db30e80-d846-11eb-8361-d0bad7ed9c94.png)

**Additional context**
N/A, small database, minimal test does not yield expected behaviour

spl:
Thanks for this, too. We're currently working on revamping a lot of things at the moment. But we'll look into this problem soon.

GavinMendelGleason:
Yes, actually this *should* be solved by #755 but the syntax will be to use '.' rather than '*' which is reserved for pattern repetition multiple times (as in Regex). It is not yet in the JS or Python client yet, but should be by next week after we upgrade the TerminusX servers.

hoijnet:
Looking forward to trying it out!

PR #755, opened 2021-11-10
Adding "any" path pattern ( '.' in regex )

GavinMendelGleason:
Required for CAMS use case


matko (approved):

matko (approved):

hoijnet:
Curious as to if this also solved the wildcard any pattern mentioned in https://github.com/terminusdb/terminusdb/issues/347

PR #764, opened 2021-11-12
Improve input checking for document GET/DELETE

spl:
* Create `src/core/util/param.pl` for checking JSON and search parameters
* Fix many issues with unchecked bad parameter types
* Add tests for document query and deletion
* Consolidate error reports under `bad_parameter_type`
* Remove unused or replaced error responses

### DIFF
--- a/src/core/util/param.pl
+++ b/src/core/util/param.pl
@@ -1,0 +1,156 @@
+/*
+ * This is a collection of utilities for finding, checking, and converting
+ * values from JSON and search (a.k.a. query string parameters).
+ *
+ * All JSON and search values should involve one of these predicates so that we
+ * can have consistent and safe input validation along with minimal redundancy.
+ */
+:- module(param, [
+              param_check_json/4,
+              param_check_search/4,
+
+              param_value_search/4,
+              param_value_search_optional/5,
+              param_value_json/4,
+              param_value_json_optional/5,
+              param_value_search_or_json/5,
+              param_value_search_or_json_optional/6,
+
+              param_value_search_author/2,
+              param_value_search_message/2,
+              param_value_search_graph_type/2
+          ]).
+
+/*
+ * Check a JSON dict parameter value against an expected type. If it's wrong,
+ * throw an error. Otherwise, convert it if necessary.
+ */
+param_check_json(Type, Param, Value_In, Value_Out) :-
+    (   param_check_json_(Type, Value_In, Value_Out)
+    ->  true
+    ;   param_check_common_(Type, Value_In, Value_Out)
+    ->  true
+    ;   throw(error(bad_parameter_type(Param, Type, Value_In), _))).
+
+param_check_json_(graph, "schema", schema).
+param_check_json_(graph, "instance", instance).
+param_check_json_(object, Value_In, Value_Out) :-
+    is_dict(Value_In),
+    Value_Out = Value_In.
+param_check_json_(string, Value_In, Value_Out) :-
+    string(Value_In),
+    Value_In \= "",
+    Value_Out = Value_In.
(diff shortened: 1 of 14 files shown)


src/core/util/param.pl
```diff
@@ -0,0 +1,122 @@
...
+param_check_json(string, _, Value_In, Value_Out) :-
+    string(Value_In),
+    Value_In \= "",
+    !,
+    Value_Out = Value_In.
+param_check_json(atom_or_string, Key, Value_In, Value_Out) :-
+    param_check_json(string, Key, Value_In, Value_String),
+    !,
+    atom_string(Value_Out, Value_String).
+param_check_json(Type, Key, Value_In, _) :-
```
GavinMendelGleason, line 51: No serious thing especially in this case but it is slightly better stylistically and in performance to make an inner predicate which is semi-det and first argument indexed on an atom, and then have an outer predicate which is det+exception which judges if the inner predicate fails or not. 

This way no cuts will be required.

GavinMendelGleason: e.g.

```prolog
param_check_json(Type, Key, Value_In, Value_Out) :- 
  (    param_check_json_(Type, Key, Value_In, Value_Out)
  ->  true
  ;    throw(error(bad_parameter_type), _)
  ) .

param_check_json_(boolean, _, false, false).
param_check_json_(boolean, _, true, true).
param_check_json_(graph, _, "schema", schema).
param_check_json_(graph, _, "instance", instance).
param_check_json_(object, _, Value_In, Value_Out) :-
    is_dict(Value_In),
    Value_Out = Value_In.
...
```

src/server/routes.pl
```diff
@@ -774,57 +774,32 @@
             ->  true
             ;   Search = []),
 
-            (   get_dict(graph_type, Posted, Graph_Type_String)
-            ->  atom_string(Graph_Type, Graph_Type_String)
-            ;   memberchk(graph_type=Graph_Type, Search)
-            ->  do_or_die(memberchk(Graph_Type, [schema, instance]),
-                          error(unknown_graph_type(Graph_Type),_))
+            (   param_value_search_or_json(Search, JSON, graph_type, graph, Graph_Type)
```
GavinMendelGleason, line 777: Fair number of choice points set up here. This increases the stack size and can create surprising behaviour if we back track over these to find the default solution which is probably not what is meant.  (this could be controlled with a cut at the end of the parameter selections but that is non-local and confusing) 

It is better to have a default value version of `param_value_search_or_json` which takes another arg and uses if-then-else. 

i.e.
```prolog
param_value_search_or_json(Search, JSON, Arg, Type, Default, Value) :- 
    (    param_value_search_or_json(Search, JSON, Arg, Type, Value) 
    ->  true
    ;    Value = Default
    ).
```

GavinMendelGleason (changes requested):
Overall much cleaner! 

The defaulty behaviour (final error) need not be addressed, but I think the choice points should be.

src/core/util/param.pl
spl: Good style. Done!

src/server/routes.pl
spl: Great! I just have to remember that `->/2` implies a cut and that `(... ; ...)` without `->/2` implies there may be unexpected backtracking.

spl:
[These](https://github.com/terminusdb/terminusdb/compare/695ecaa75e3b22d5e4ec27d57eafc965184794c9..249ea6b29f7aed949d2fa279f16e99db875914ff) are the recent changes (minus the version bump introduced in `main`).

GavinMendelGleason (approved):

PR #768, opened 2021-11-15
Use local version of terminusdb, not from the full path

GavinMendelGleason:


spl:
This change only works if you're running the script from the top-level repository directory. Could we instead move the body of the script into the `docs` target of the `Makefile` and remove the script?

GavinMendelGleason:
Absolutely - this is evidently the reason that the result was empty in the md file.

PR #770, opened 2021-11-15
add generation for all frames

### LINKED ISSUE #767: implement a call which returns all frames irrespective of type
**Describe the bug**
implement a call which returns all frames irrespective of type. So I could easily get frames of a subdocument as well.

GavinMendelGleason:
fixes #767


spl:
Can you add tests and use the `param.pl` module introduced by #764 once it is merged?

matko (approved):

PR #778, opened 2021-11-16
Add tests for `prefixed`

spl:
* Add tests for the `prefixed` parameter for `graph_type=instance`
* Rename `Prefix` to `Compress` in `api_document.pl`

`prefixed` does not currently work with `graph_type=schema`. That issue is tracked in #801.

### DIFF
--- a/src/server/routes.pl
+++ b/src/server/routes.pl
@@ -779,7 +779,7 @@ document_handler(get, Path, Request, System_DB, Auth) :-
             param_value_search_or_json_optional(Search, JSON, count, nonnegative_integer, unlimited, Count),
             param_value_search_or_json_optional(Search, JSON, minimized, boolean, true, Minimized),
             param_value_search_or_json_optional(Search, JSON, as_list, boolean, false, As_List),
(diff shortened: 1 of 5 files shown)


src/server/routes.pl
```diff
@@ -790,17 +790,22 @@
             ->  JSON_Options = [width(0)]
             ;   JSON_Options = []),
 
+            % Compress is the logical negation of Prefixed.
```
matko, line 793: It really isn't. prefixed means we turn a long string 'http://foo/bar' into 'foo:Bar'. So that is actually a compression. For the cases where something lives under `@base` or `@schema`, in the proper contexts, we can completely omit the prefix, because it is the default.

matko (changes requested):

src/server/routes.pl
spl: Thanks. I reverted that change. But I kept the rename of `Prefixed` to `Compress` because it was confusing to have different arg names in different files.

GavinMendelGleason (approved):

matko (approved):

PR #793, opened 2021-11-18
Query type fixes

matko:
It turns out that we were not properly considering the specified type in the query interface, occasionally returning too many objects. Also, a subsumption check was wrong.


GavinMendelGleason (approved):

GavinMendelGleason:
To check this, create a class hierarchy in the schema. Then see if you get all documents below the super-class. Check if you only get documents from the subclass. 

Also we should check set, optional search.

PR #821, opened 2021-11-30
Adding some fixes for WOQL treatment of documents

GavinMendelGleason:


src/core/query/woql_compile.pl
```diff
@@ -1546,7 +1542,7 @@
     compile_arith(Arith,Pre_Term,ArithE).
 compile_wf(dot(Dict,Key,Value), get_dict(KeyE,DictE,ValueE)) -->
     resolve(Dict,DictE),
-    resolve(Key,KeyE),
+    {atom_string(KeyE,Key)},
```
spl, line 1545: What is the difference between `{atom_string(KeyE,Key)}` and `atom_string(KeyE,Key)`?

spl:
Do we need to update the clients for this change?

src/core/query/woql_compile.pl
GavinMendelGleason: We are in a DCG so there are two implicit parameters unless we wrap in a `{ }` block

GavinMendelGleason:
Yes, but I don't think there is any regression as the clients are not correctly implementing the document interface in WOQL yet. We need to create integration tests in both.

Issue #825, opened 2021-12-01
Text encoding issues in document query


spl:
Consider the following script:

```sh
#!/bin/bash
set -ex

# Delete and create a new database
xh DELETE 'http://admin:root@localhost:6363/api/db/admin/tdb' || true
xh 'http://admin:root@localhost:6363/api/db/admin/tdb' label=l comment=c

# Insert the schema
xh 'http://admin:root@localhost:6363/api/document/admin/tdb' graph_type==schema author==a message==m <<EOF
{"@id":"X","@type":"Class","@key":{"@type":"Lexical","@fields":["s"]},"s":"xsd:string"}
EOF

# Insert an instance with a non-ASCII string
xh 'http://admin:root@localhost:6363/api/document/admin/tdb' author==a message==m <<EOF
{"@type":"X","s":"Kurt Gödel"}
EOF

# Fetch the instances
xh 'http://admin:root@localhost:6363/api/document/admin/tdb'
```

The output from the last command is:

```json
{
    "@id": "X/Kurt%20G%C3%83%C2%B6del",
    "@type": "X",
    "s": "Kurt GÃ¶del"
}
```

Observations:
* `@id` is URL-encoded. Does it need to be for JSON? It means more work comparing the value passed in.
* `s` is ugh.

spl:
Similar output with `curl`:

```sh
#!/bin/bash
set -ex

# Delete and create a new database
curl -X DELETE 'http://admin:root@localhost:6363/api/db/admin/tdb'
curl -X POST -H 'Content-Type: application/json' 'http://admin:root@localhost:6363/api/db/admin/tdb' -d @- <<EOF
{"label":"l","comment":"c"}
EOF

# Insert the schema
curl -X POST -H 'Content-Type: application/json' 'http://admin:root@localhost:6363/api/document/admin/tdb?graph_type=schema&author=a&message==m' -d @- <<EOF
{"@id":"X","@type":"Class","@key":{"@type":"Lexical","@fields":["s"]},"s":"xsd:string"}
EOF

# Insert an instance with a non-ASCII string
curl -X POST -H 'Content-Type: application/json' 'http://admin:root@localhost:6363/api/document/admin/tdb?author=a&message=m' -d @- <<EOF
{"@type":"X","s":"Kurt Gödel"}
EOF

# Fetch the instances
curl 'http://admin:root@localhost:6363/api/document/admin/tdb'
```

Output of the last command:

```json
[...]

PR #828, opened 2021-12-03
Update terminus_store_prolog to v0.19.5

### LINKED ISSUE #637: Upgrade to swipl 8.4
swipl 8.4 has just been released. Let's try to upgrade to it.

spl:
* Updates `swipl` to v8.4.1
* Closes #637


spl:
I initially wanted this for fixing #825, but I figured out a workaround in #833, so this is not needed immediately.

spl:
@matko said in the past that the `connect_handler` tests were hanging for him with `swipl` v8.4.0. At the time, I don't think they were hanging for me, but now they are.

Looking into one of the tests, I broke it into two parts. The first doesn't hang in `make i`:

```prolog
test_utils:setup_temp_server(Store-Dir-PID, Server),
test_utils:admin_pass(Key),
atomic_list_concat([Server, '/api/'], URL),
http_client:http_get(URL, _, [authorization(basic(admin, Key))]),
process_kill(PID).
```

The second hangs:

```prolog
process_wait($PID, _).
```

spl:
If I use `process_kill(PID, 9)` instead of `process_kill(PID)`, then `process_wait($PID, _)` does not hang.

spl:
This makes me think it's a change in the way the `swipl` process can be killed. The default sent by `process_kill/1` is `SIGTERM` (15). But, for some reason, this is not killing the `swipl` process. On other hand, `SIGKILL` (9) does kill it.

spl:
Reported this issue at <https://github.com/SWI-Prolog/swipl/issues/19>.

matko (approved):
Unit tests that spawn extra servers run for me now - great!

PR #831, opened 2021-12-06
Introduce mock header

rrooij:
Some systems can be setup in such a way to offload user authentication. For instance, you could set up TerminusDB in a protected environment and use oauth2-proxy to secure it.

Therefore we now allow an insecure mode to be set in conjunction with a custom user header to authenticate.

The ENV variables required for this functionality are:

TERMINUSDB_INSECURE_MODE which has to be set to a value and TERMINUSDB_FORWARDED_USER_HEADER with the name of the HTTP header which contains the username.

### DIFF
--- a/src/server/routes.pl
+++ b/src/server/routes.pl
@@ -146,8 +146,8 @@ test(connection_authorised_user_http_basic, [
 
 test(connection_authorised_user_forwarded_header, [
          setup(setup_temp_server(State, Server, [env_vars([
-                                     'TERMINUSDB_INSECURE_MODE'=1,
-                                     'TERMINUSDB_FORWARDED_USER_HEADER'='X-Forwarded-User'
+                                     'TERMINUSDB_INSECURE_USER_HEADER_ENABLED'=true,
+                                     'TERMINUSDB_INSECURE_USER_HEADER'='X-Forwarded-User'
                                  ])])),
          cleanup(teardown_temp_server(State))
      ]) :-
@@ -160,8 +160,8 @@ test(connection_authorised_user_forwarded_header, [
 
 test(connection_unauthorised_user_forwarded_header, [
          setup(setup_temp_server(State, Server, [env_vars([
-                                     'TERMINUSDB_INSECURE_MODE'=1,
-                                     'TERMINUSDB_FORWARDED_USER_HEADER'='X-Forwarded-User'
+                                     'TERMINUSDB_INSECURE_USER_HEADER_ENABLED'=true,
+                                     'TERMINUSDB_INSECURE_USER_HEADER'='X-Forwarded-User'
                                  ])])),
          cleanup(teardown_temp_server(State))
      ]) :-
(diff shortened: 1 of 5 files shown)


hoijnet:
I would suggest to add in the documentation that the recommended way is to do pass-through authentication and use the JWT token to authenticate also in the backend rather than to use the insecure header way :-)

rrooij:
> I would suggest to add in the documentation that the recommended way is to do pass-through authentication and use the JWT token to authenticate also in the back-end rather than to use the insecure header way :-)

Thanks for taking a look. :smile: I agree that we need to document the correct way to do this! But... if you have pass-through authentication anyway, using the JWT in TerminusDB does not make much sense anymore, right? For instance, you could use another service (like oauth2-proxy) to solely do the authentication part.

So in my opinion, the best flow would be:

Some API gateway doing authentication or a reverse proxy with oauth2-proxy and the like -> pass the user header (instead of the JWT)

hoijnet:
The insecure header-based authentication hinges on thew assumption that no other services can make unauthenticated network calls to TerminusDB, neither TerminusDB, Oauth2-proxy or any other service. Whilst that would expected to be the case in normal operations, in an attack, one or more of the individual services would get compromised and could potentially send an authenticated network call to self or to the services around, knowing which service to impersonate, and the header to use.

Whilst quite a lot more secure to have a proxy authenticating in front and then passing on the request, passing through the JWT solves to secure against mistakes too. I would say they are different levels of security, and that many would be satisfied with the lower level. Sometimes passing a JWT to the backend is not an option, and then a header-based authorisation could be the only option.

The simple example where the passthrough would become a significant issue is a configuration change of the frontend where the header is not protected (the client may write it itself), that the proxy is accidentally put into passthrough in some environment, or other similar issues.

I believe the recommendation, if using JWKS- and JWT-based authentication, should be to pass the JWT to backend services for authentication services, as long as the JWT can be held reasonably secure in them (such as in the case with TerminusDB). Suggesting that to be the recommendation.

hoijnet:
I'm not against having it, just to ensure the risks are documented so that admins make the choice conciously =)

spl (changes requested):
Most of my comments are around naming consistency. I've suggested some names, but I'm open to alternatives, as long as we use them consistently.

src/server/routes.pl
```diff
@@ -144,6 +144,33 @@
...
 
 
+test(connection_authorised_user_forwarded_header, [
+         setup(setup_temp_server(State, Server, [env_vars([
+                                     'TERMINUSDB_INSECURE_MODE'=1,
+                                     'TERMINUSDB_FORWARDED_USER_HEADER'='X-Forwarded-User'
+                                 ])])),
+         cleanup(teardown_temp_server(State))
+     ]) :-
+    create_db_without_schema("admin", "TEST_DB"),
```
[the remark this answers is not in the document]
rrooij: I implemented most changes you suggested. About the ENV variables in the test, I think my way of doing this is cleaner and results in less boilerplate than using `setenv`. On some operating systems, not all ENV variables are inherited automatically as well.

src/server/routes.pl
```diff
@@ -3675,23 +3675,23 @@
                        user: Username
                    }).
 authenticate(System_Askable, Request, Auth) :-
-    user_forwarded_header(Functor),
-    Term =.. [Functor, Username],
+    insecure_user_header(Header_Key),
+    Term =.. [Header_Key, Username],
```
spl, line 3679: What happens if `Header_Key` is...?
* `''`
* `' '`
* `_`
* `.`
* something else I'm probably not thinking of

Looking at `=../2` in `swipl`, it seems to just use these as atoms directly. However, I don't think it would be good to rely on that, and I don't think we should expect any values like these.

Should there be restrictions on what `TERMINUSDB_FORWARDED_USER_HEADER` can be? I think it would be reasonable to accept a **string matching the regular expression `[A-Za-z0-9\-]+`**. (Note that it must be non-empty due to `+`.) We can always expand on that if, for some reason, it is needed.

spl (changes requested):

src/server/routes.pl
rrooij: That makes sense

rrooij: Swipl complains already when you have very weird HTTP headers, example with single quotes containing header key:

```

Illegal HTTP parameter: X-'Forwarded User': admin (in_http_request)</p>
```

rrooij:
Implemented a check at startup, since I found it a waste of resources to check the header name at each request and harder for server administrators to debug.

spl:
I really like the check at startup, but wouldn't it be better to keep the value at that point and pass it down through options instead of having 2 uses of `getenv('TERMINUSDB_INSECURE_USER_HEADER', Value)`? I can just imagine someone renaming the variable name in one place and not another... 😱

rrooij:
I do think the chance of this is very, very small since it is in the same small source file.

rrooij:
Fixed it anyway :smile:

spl:
That's not quite what I was thinking. I was imagining using `getenv('TERMINUSDB_INSECURE_USER_HEADER', Value)` once and passing or storing that `Value` as an option such that the route handler could use. I am not aware of existing code that does this, however.

Something else occurred to me. (Sorry! I didn't think of this earlier.) The current placement of `check_env_vars` means that it won't be called if we run `terminus_server` in the REPL. Also, considering there is already configuration being done in `terminus_server`, I think it would make more sense there.

rrooij:
I was thinking of placing the ENV var check in the main server predicate, but we might introduce command line options that need certain ENV variables as well. For instance, what if we want to check ENV variables for `terminusdb list`?

rrooij:
Implemented it in the `terminus_server` predicate. :smile:

PR #832, opened 2021-12-07
Fix target path for building docs

spl:
This allows us to build the docs without worrying about the target path.


rrooij (approved):
Great! The target_dir oneliner is a bit messy, but that's the way makefiles work often unfortunately. 

And indeed, the target could be somewhere else than the root of the Makefile dir. Makes sense!

PR #836, opened 2021-12-09
Add insecure user header for authentication

spl:
* test_utils: add option for env vars on spawn

  Sometimes a test needs additional ENV vars even though it might not
  be necessary to set them globally for every test. Therefore I
  introduced the extra option called `env_vars`.

* Add authentication with insecure header for username

  Some systems can be setup in such a way to offload user
  authentication. For instance, you could set up TerminusDB in a
  protected environment and use oauth2-proxy to secure it.

  Therefore we now allow an insecure mode to be set in conjunction with
  a custom user header to authenticate.

* Implement env vars with dynamic predicates and robust checking for
  insecure header

* Check the env vars at initialization

* Closes #831

### DIFF
--- a/src/cli/main.pl
+++ b/src/cli/main.pl
@@ -561,6 +561,8 @@ command_subcommand(Command,Subcommand) :-
     opt_spec(Command,Subcommand,_,_,_).
 
 run(Argv) :-
+    % Check env vars to report errors as soon as possible.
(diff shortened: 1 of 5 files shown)


rrooij (approved):
Reviewed it, tried it locally and works exactly as intended. Thanks a lot!

PR #872, opened 2022-01-07
routes: fix stuck connection on empty post data

### LINKED ISSUE #864: Missing document in DB API does not close connection
**To Reproduce**
1. Start the server
2. Try the curl command: `curl localhost:6363/api/db/admin/MyNewDatabase -u admin:root -X POST -H 'Content-Type: application/json'`

**Expected behavior**
A closed connection and an error, instead of keeping the connection open. It works correctly when I supply an empty document:

```
curl localhost:6363/api/db/admin/MyNewDatabase -u admin:root -X POST -H 'Content-Type: application/json'` -d '{}'
```

**Information**

Using the latest dev docker

rrooij:
If there is no content, http_read_data will get stuck until the connection times out. Therefore I check whether there is some content first.

Fixes #864

### DIFF
--- a/src/server/routes.pl
+++ b/src/server/routes.pl
@@ -3956,6 +3956,7 @@ add_payload_to_request(Request,[payload(Document)|Request]) :-
     !.
 add_payload_to_request(Request,[payload(Document)|Request]) :-
     memberchk(content_type(_Some_Other_Type), Request),
+    check_content_length(Request),
     !,
     http_read_data(Request, Document, []).
 add_payload_to_request(Request,Request).


spl:
We didn't have enough time to cover this earlier, but why is it better to `throw` (which `check_content_length` does) than `fail` here?

spl:
Also, it seems like it should be very easy to put an integration test together for this. Can we do that?

rrooij:
Another problem with this solution, it keeps a connection open when you set the Content-Length manually:

```
curl localhost:6363/api/db/lmao/test -X POST -u admin:root -H 'Content-Type: application/json' -H 'Content-Length: 10'
```

spl:
@rrooij Can you change the `check_content_length` to `memberchk` and fail? If it works for you and doesn't cause any other problems, we'll go with that.

matko:
I disagree with this solution.
@rrooij as you point out already, we are still dealing with a hanging request in case that the header is supplied but then no data follows. The real problem here is that there's the possibility to open a connection to us, and just keep the connection open, not submitting any data, thereby locking up a worker thread. Your change does not mitigate this.

It is actually not illegal to send a request where you don't specify the content-length up front. Often times, when generating data, a client process may simply not know in advance how much data there's going to be. Forcing them to specify a content-length means forcing them to first buffer all data locally, and then send it in one go. Not requiring the header means they can just submit as they generate data.

Since this change doesn't actually fix the problem, but does break a valid use case, I'm against merging this.

PR #874, opened 2022-01-10
bootstrap: import plunit as library

rrooij:
We use set_test_options which is a predicate defined in plunit. I do think we should avoid to rely on autoloading as much as possible.

### DIFF
--- a/src/bootstrap.pl
+++ b/src/bootstrap.pl
@@ -9,6 +9,7 @@
 :- reexport(core(util/syntax)).
 
 :- use_module(core(util)).
+:- use_module(library(plunit)).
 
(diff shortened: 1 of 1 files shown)


spl:
I agree. I've gotten caught by this before. Can we disable it? Would [`autoload(explicit)`](https://www.swi-prolog.org/pldoc/man?section=flags#flag:autoload) do the desired thing?

rrooij:
We tried to disable autoloading entirely, but unfortunately, even a lot of built-in SWI Prolog modules make extensive use of it. For instance, there are several unimported modules autoloaded by `qsave` itself.

spl (approved):

PR #876, opened 2022-01-10
Import lists explicitly where member/2 is called

rrooij:
The member/2 predicate is used extensively throughout the code but the module where it comes from is almost never loaded.

This commit fixes that by importing it in every file where member/2 is used.


GavinMendelGleason (approved):
Should be safe enough :D

PR #878, opened 2022-01-11
Id capture

matko:
This pull request introduces id captures, a feature of the document interface where, when one does an insert or replace operation, they can capture ids, and refer to them elsewhere in the same insert/update stream.

This supports both backwards references (where you first capture an id, then refer to it later), and forward references (where you first refer to an id, then capture it later).

### DIFF
--- a/src/core/triple/triplestore.pl
+++ b/src/core/triple/triplestore.pl
@@ -231,6 +231,8 @@ import_graph(_File, _DB_ID, _Graph_ID) :-
  * Insert triple into transaction layer, record changed as 1 or 0
  */
 insert(G,X,Y,Z,Changed) :-
+    do_or_die(ground(Z),
+              error(instantiation_error, _)),
     ground_object_storage(Z,S),
     read_write_obj_builder(G, Builder),
     (   nb_add_triple(Builder, X, Y, S)
--- a/src/core/api/api_document.pl
+++ b/src/core/api/api_document.pl
@@ -225,6 +225,7 @@ api_insert_documents(SystemDB, Auth, Path, Schema_Or_Instance, Author, Message, 
 
     stream_property(Stream, position(Pos)),
 
+    ensure_transaction_has_builder(instance, Transaction),
     with_transaction(Context,
                      (   set_stream_position(Stream, Pos),
                          Full_Replace = true
--- a/src/core/api/api_document.pl
+++ b/src/core/api/api_document.pl
@@ -359,14 +359,11 @@ api_replace_documents(SystemDB, Auth, Path, Schema_Or_Instance, Author, Message,
                                      ),
                                      Captures_Var),
                                  Ids),
-                         json_log_info_formatted("done replacing", []),
                          die_if(nonground_captures(Captures_Var, Nonground),
                                 error(not_all_captures_found(Nonground), _)),
-                         die_if(has_duplicates(Ids, Duplicates), error(same_ids_in_one_transaction(Duplicates), _)),
-                         json_log_info_formatted("awafsd", [])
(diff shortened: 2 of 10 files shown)


matko:
I've set up auto-merge as a squash commit, since there's a bunch of low quality intermittent commits in this pull.

src/core/triple/triplestore.pl
```diff
@@ -231,6 +231,8 @@
  * Insert triple into transaction layer, record changed as 1 or 0
  */
 insert(G,X,Y,Z,Changed) :-
+    do_or_die(ground(Z),
+              error(instantiation_error, _)),
```
spl, line 235: I've started using the following for similar cases:

```prolog
write_data_version_header(Data_Version) :-
    throw(error(unexpected_argument_instantiation(write_data_version_header, Data_Version), _)).
```

Is there some error like `unexpected_argument_instantiation(insert, Z)` that would be more descriptive in identifying this problem when it appears?

GavinMendelGleason: This is what prolog throws in libraries according to a standard convention. I agree it is not ideal, but we'd have to catch two different cases if we don't do it this way.

spl: Do we catch `instantiation_error`? I don't see where, and I don't understand why we would. It indicates a programmatic error, not a problem with the input.

GavinMendelGleason: We do, and it tracks precisely the cases where a predicate is called with some uninstantiated variables which can occur in WOQL when using WOQL words which expect bound arguments in certain modes which is why we should throw the same error condition as we get from library predicates which are called in the wrong mode.

spl: Okay, so this is specific to `triplestore.pl` and WOQL. I thought you meant we were catching this at the route handlers.

matko:
So what is going on in this pull request?

The bulk of the change is in our elaboration process, which takes input json and turns it into an elaborated (json-ld) document. Two new keywords have been introduced, `@capture` and `@ref`. `@capture` captures the ID variable with a capture string that can be used to refer to it.. `@ref` instructs the elaborator to unify an ID reference with that variable.

Essentially, `json_elaborate` now gets 3 extra arguments, `+Captures_In`, `-Dependencies` and `-Captures_Out`, which get threaded through the entire elaboration process. the two `Captures` are an association list, one input, one output, associating a capture string with a variable.
Whenever either an `@capture` or an `@ref` is encountered, `Captures_In` is checked for an existing entry. If it's there, retrieve it. if it's not there, We make a new entry. 
In the case of `@capture`, we then ensure that the id of the document this property occured in is unified with the entry in the captures association list. The resulting elaborated json-ld document no longer contains the `@capture` property.
 In the case of `{"@ref": "Some_String"}`, we replace it altogether with `{"@type": "@id", "@id": ..whatever_we_retrieved..}`.

`Dependencies` keeps track of any encountered ref variables, so that the caller of `json_elaborate` can know what this document elaboration depends on, and if all dependencies have been resolved yet. In both `insert_document` and `replace_document`, we use `when` coroutining to delay the document insertion until the time that the dependency list is ground. This will happen when all documents whose ids we depend on have been elaborated.

the top level api call, which calls `insert_document` and `replace_document` multiple times, ensures that the captures variable is properly threaded through, starting with an empty list and ending with an association list containing every single dependency of all the documents. A final check is done to see if the capture output is ground. If it is not ground, that means that some of the documents were actually not inserted, because they're still waiting for a dependency to be resolved. This results in an error being reported back to the user.

GavinMendelGleason (approved):

PR #880, opened 2022-01-12
Autoloading issue

rrooij:
A lot of missed modules are now imported. autoload is now also enabled for qsave.


spl (approved):
I added a `make` target to check imports with autoload disabled to catch this issue in the future, and it's called during CI. I'm happy with this PR.

PR #895, opened 2022-01-13
Check the arguments to create_context

spl:
Prevent stupid bugs from happening.

### DIFF
--- a/src/core/query/ask.pl
+++ b/src/core/query/ask.pl
@@ -173,15 +173,29 @@ askable_prefixes(Transaction_Object,Prefixes) :-
 askable_prefixes(Descriptor,Prefixes) :-
     collection_descriptor_prefixes(Descriptor, Prefixes).
 
-create_context(Layer, Context) :-
+/**
+ * create_context(+Askable, -Context) is semidet.
+ *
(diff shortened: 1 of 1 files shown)


src/core/query/ask.pl
```diff
@@ -173,15 +173,28 @@
...
+ *
+ * Create a context. Fail if Askable does not exist.
+ */
+create_context(Askable, Context) :-
+    ground(Askable),
+    var(Context),
+    !,
+    create_context_(Askable, Context).
+create_context(Askable, Context) :-
+    throw(error(unexpected_argument_instantiation(create_context, Askable, Context), _)).
```
matko, line 187: create_context_ may fail for various reasons, many of which have nothing to do with the groundless or varness of the input/output. Therefore fallthrough case catches too much.
I think it'd be better to use a `die_if` with the ground/var checks.

matko (changes requested):

src/core/query/ask.pl
spl: You're right. I had that thought before I started, but I lost it somewhere in the middle.

matko (approved):

PR #901, opened 2022-01-18
Have a rust module as part of the build for foreign predicates

matko:
Sometimes we wish to add a foreign rust predicate for speed reasons. However, so far to do so we had to write an external library which we then had to `pack_install`. In practice, this meant we extended `terminus_store_prolog` with stuff that shouldn't really be there, as it wasn't related to store at all. Furthermore, this leads to possible version mismatches, since only one pack version can be installeed at once, but different versions of terminusdb may be expecting different foreign predicates with different contracts to be there.

This pull request adds a rust module directly in the terminusdb source and includes it in the build. This allows us to add any such foreign predicates directly to the terminusdb source.


spl:
This is great! Works for me on macOS.

Makefile
```diff
@@ -58,13 +73,18 @@ docs-clean:
...
-$(TARGET):
+$(TARGET): $(RUST_TARGET)
 	# Build the target and fail for errors and warnings. Ignore warnings
 	# having "qsave(strip_failed(..." that occur on macOS.
 	$(SWIPL) -t 'main,halt.' -O -q -f src/bootstrap.pl 2>&1 | \
 	  grep -v 'qsave(strip_failed' | \
 	  (! grep -e ERROR -e Warning)
 
+$(RUST_TARGET): $(RUST_FILES)
+	echo $(RUST_FILES)
```
spl, line 84: Is this for debugging in the future?

matko: no, that's an accidental echo I left in :). I'll remove it.

spl (approved):
rustacean

PR #907, opened 2022-01-19
Fix makefile targets to support incremental build

matko:
Some more Makefile changes
- I changed the doc build to use a relative path rather than an absolute path. As far as I can tell, using ./terminusdb works just as well as using the absolute path to terminusdb, and it saves us from having to do the magic to find the absolute path. This also means that make is once again able to detect that terminusdb has been built, which it couldn't do anymore due to mild path mangling in the absolute path target.
- I added all prolog, json, and ttl files in our source dir as a dependency to the terminusdb target. This means that running make will only rebuild terminusdb when one of these files has changed since the last build.


spl (approved):
🚀

PR #911, opened 2022-01-21
Add data versions for document and woql endpoints

### LINKED ISSUE #248: Add commit ID to API requests and responses
Operation:
* The server will look for a certain HTTP header.
  * If the header is present, the server compares the commit ID with the latest commit ID. If these are different, the server responds with an error.
  * If the header is not present, do nothing.
* The server will include an HTTP header with the latest commit ID in the response.

Endpoints (`/api/...`):
* `document`
* `woql`
  * can query and update multiple databases (e.g. with `Using`)
  * use primary database only for now
* others in the future

Notes
* system and metadata need layer ID

spl:
* Read TerminusDB-Data-Version header
* Check it with the transaction data version and fail for mismatch
* Respond with possibly updated TerminusDB-Data-Version header
* Fixes #248

### DIFF
--- a/src/core/util/data_version.pl
+++ b/src/core/util/data_version.pl
@@ -3,7 +3,7 @@
               read_data_version_header/2,
               write_data_version_header/1,
               transaction_data_version/2,
-              transaction_data_version/3,
+              validation_data_version/3,
               meta_data_version/3
           ]).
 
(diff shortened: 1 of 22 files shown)


src/core/util/data_version.pl
```diff
@@ -0,0 +1,180 @@
...
+            ;   validation_object{} :< Object
+            )),
+        error(unexpected_argument_instantiation('transaction_data_version/3', objects(Object, Objects)), _)),
+    Descriptor = Object.descriptor,
+    (   transaction_data_version_(Descriptor, Object, Data_Version)
+    ->  true
+    ;   get_dict(repository_descriptor, Descriptor, Repo_Descriptor),
+        % Search Objects for the matching repository object.
+        member(Repo_Object, Objects),
+        get_dict(descriptor, Repo_Object, Repo_Descriptor),
```
matko, line 135: after we found a match we should cut, ensuring that no choice point is left.

src/core/util/data_version.pl
```diff
@@ -0,0 +1,180 @@
...
+    transaction_data_version_(Object.descriptor, Repo_Object, Data_Version).
+
+/**
+ * transaction_data_version(+Object, +Objects, -Data_Version) is semidet.
+ *
+ * Look for the data version of a transaction/validation object using Objects to
+ * look for Object's repository transaction/validation object. Fail if Object
+ * does not have a data version.
+ */
+transaction_data_version(Object, Objects, Data_Version) :-
```
matko, line 121: It looks like this case is used for extracting the data version from validation objects only, as for transaction objects we can always just use the parent transaction.
Maybe that can be clarified in a comment.

matko (changes requested):
Well done, this will work just fine. I'm just requesting a small change to add a cut at some point.

src/core/util/data_version.pl
spl: Nice one!

src/core/util/data_version.pl
spl: Good point. Maybe we should even go further?

1. Rename `transaction_data_version/3` to `validation_data_version/3`
2. Remove unnecessary checks:
   * `transaction_object{}` from `validation_data_version/3`
   * `validation_object{}` from `transaction_data_version/2`

What do you think?

matko: Yeah, that's a good idea. We won't be able to accidentally use it in a wrong way then.

src/core/util/data_version.pl
spl: https://github.com/terminusdb/terminusdb/pull/911/commits/d8bc2413a5d6e5d4800afbf1322d3e60239459fa

src/core/util/data_version.pl
spl: https://github.com/terminusdb/terminusdb/pull/911/commits/dc28a52379efcf8b9926961dcbddabc6dd1e6120

spl:
This is ready for a follow-up review.

matko (approved):

PR #928, opened 2022-01-26
Missing autoload

rrooij:
A lot of autoloading fixes. I will try to make the check better to prevent this from happening.


src/core/query/metadata.pl
```diff
@@ -23,7 +23,7 @@
 transaction_object_triple_count(Transaction, Triple_Count) :-
     transaction_object_read_write_objs(Transaction,Objects),
     maplist(read_object_triple_count,Objects,Counts),
-    sumlist(Counts,Triple_Count).
+    sum_list(Counts,Triple_Count).
```
spl, line 26: Why is this change necessary? Are these predicates containing `sumlist` not being used?

spl (changes requested):

spl:
> I will try to make the check better to prevent this from happening.

Can you test that with and without these changes and add it to the PR? I think it would be ideal to avoid making changes like this unless we can prevent regressions.

src/core/query/metadata.pl
rrooij: It couldn't autoload sumlist but I could not find the library that it would need to be imported from. Therefore I changed it to the non-deprecated one since the predicate signature is the same.

rrooij:
I solved all the remaining autoloaded libraries and implemented a very simple linter in Python.

.github/workflows/ci.yml
```diff
@@ -91,7 +101,7 @@ jobs:
       - name: Run server
         run: |
           docker load < terminusdb-server-docker-image.tar.gz
-          docker run $DOCKER_IMAGE_NAME:local make check-imports
+          docker run $DOCKER_IMAGE_NAME:local make
```
spl, line 104: I just realized: this line should go. We don't need `make` here.

spl (changes requested):

rrooij:
I think everything is resolved now. We can look for a cleaner solution in the future. :smile:

spl (approved):

PR #937, opened 2022-01-28
Table type implementation

GavinMendelGleason:
It would be nice to have Nth Order lists, and actually this isn't very hard.

### DIFF
--- a/src/core/document/instance.pl
+++ b/src/core/document/instance.pl
@@ -246,6 +246,38 @@ refute_cardinality_(list,Validation_Object,S,P,Witness) :-
                        instance: Subject,
                        predicate: Predicate
                      }.
+refute_cardinality_(table(C),Validation_Object,S,P,Witness) :-
+    % a property whose value is a list
+    \+ card_count(Validation_Object,S,P,1),
+    instance_layer(Validation_Object, Layer),
+    terminus_store:subject_id(Layer, Subject_String, S),
+    atom_string(Subject, Subject_String),
+    (   atom(P)
+    ->  P = Predicate
+    ;   terminus_store:predicate_id(Layer, Predicate_String, P),
+        atom_string(Predicate, Predicate_String)
+    ),
+    test_utils:print_all_triples(Validation_Object, schema),
+    Witness = witness{ '@type': instance_not_cardinality_one,
+                       instance: Subject,
+                       class: C,
+                       predicate: Predicate
+                     }.
+refute_cardinality_(table,Validation_Object,S,P,Witness) :-
+    % a property inside a list cell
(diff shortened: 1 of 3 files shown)


src/core/document/instance.pl
```diff
@@ -246,6 +246,38 @@
...
+    \+ card_count(Validation_Object,S,P,1),
+    instance_layer(Validation_Object, Layer),
+    terminus_store:subject_id(Layer, Subject_String, S),
+    atom_string(Subject, Subject_String),
+    (   atom(P)
+    ->  P = Predicate
+    ;   terminus_store:predicate_id(Layer, Predicate_String, P),
+        atom_string(Predicate, Predicate_String)
+    ),
+    test_utils:print_all_triples(Validation_Object, schema),
```
matko, line 260: we probably don't want this.

src/core/document/json.pl
```diff
@@ -704,11 +731,34 @@
...
 context_value_expand(_,_,null,_,Captures,null,[],Captures) :-
     !.
+context_value_expand(DB,Context,Value,Expansion,Captures_In,V,Dependencies,Captures_Out) :-
+    get_dict('@container', Expansion, "@table"),
+    !,
+    % Container type
+    get_dict('@type', Expansion, Elt_Type),
+    (   is_list(Value)
+    ->  Value_List = Value
+    ;   string(Value)
```
matko, line 752: what does this case mean for a table?

src/core/document/json.pl
```diff
@@ -704,11 +731,34 @@
...
+    ->  Value_List = [Value]
+    ;   get_dict('@value',Value,Value_List)
+    ->  true
```
matko, line 755: here too - is this a copy of the list code? are we sure that just works for tables?

src/core/document/json.pl
```diff
@@ -1484,6 +1537,23 @@
...
+    idgen_random(Base_Cons,New_ID),
+    (   Triple = t(ID,Key,New_ID)
+    ;   global_prefix_expand(rdf:type, RDF_Type),
+        global_prefix_expand(rdf:'List', RDF_List),
+        Triple = t(New_ID, RDF_Type, RDF_List)
+    ;   global_prefix_expand(rdf:first, RDF_First),
+        list_id_key_context_triple(H,New_ID,RDF_First,Context,Triple)
+    ;   global_prefix_expand(rdf:rest, RDF_Rest),
+        table_id_key_context_triple(T,New_ID,RDF_Rest,Context,Triple)
+    ;   json_triple_(H,Context,Triple)
```
matko, line 1554: this case is bad maybe

matko (changes requested):

matko (approved):

PR #942, opened 2022-02-01
info: add unknown version of terminusdb-store if it isn't found

rrooij:
The pack_property predicate fails a lot when specific things aren't set up the way it expects. For instance, you have to name your pack directory *exactly* the same as the pack name even though this is not required to use packs at all.

### DIFF
--- a/src/core/api/api_info.pl
+++ b/src/core/api/api_info.pl
@@ -10,7 +10,9 @@ info(_System_DB, Auth, Info) :-
         error(access_not_authorized(Auth),_)),
 
     version(TerminusDB_Version),
-    pack:pack_property(terminus_store_prolog, version(TerminusDB_Store_Version)),
+    (   pack:pack_property(terminus_store_prolog, version(TerminusDB_Store_Version))
+    ->  true
+    ;   TerminusDB_Store_Version = 'unknown'),
     get_db_version(Storage_Version),
 
     Info = _{


spl:
The fact that `pack_property` fails silently is not good. But I'm not convinced this approach is better than throwing an exception saying the `terminus_prolog_store` pack is not found.

rrooij:
`terminus_store_prolog` is found when this occurs. Otherwise `use_module(library(terminus_store))` would not work and it would throw an exception for not finding the module. The only thing that it can't find is the version.

This predicate will never work with binaries compiled with `qsave`.

spl:
> `terminus_store_prolog` is found when this occurs.

But it's not. Because `pack_property` is looking for the _directory_ with the name `terminus_store_prolog`, right? That's why it fails.

> Otherwise `use_module(library(terminus_store))` would not work and it would throw an exception for not finding the module.

Yes, I understand that the pack is loaded, but the pack directory cannot be found.

> This predicate will never work with binaries compiled with `qsave` since it will never be able to find it in the right directory except when you have a SWI Prolog environment with all the packs correctly installed (which defeats the purpose of the compiled binary).

I might have a solution to that. See <https://github.com/terminusdb/terminusdb/commit/1e4cabd89d7b40b9252b914b32ed029cb71ea35d>. I probably need to add the `set_prolog_flag` part in more places.

rrooij:
That is a very elegant solution for me. :smile: 

Be wary indeed that it has to be in more places for it to work correctly, it now only works on compilation but should probably also be in `start.pl` and `interactive.pl`

matko:
As there's still ongoing discussion and an alternative solution strategy, I'm turning this into a draft.

PR #945, opened 2022-02-01
Timeout on large table diffs

GavinMendelGleason:


src/core/api/api_error.pl
```diff
@@ -1087,6 +1087,13 @@
                               'api:key' : Key},
              'api:message' : Msg
             }.
+api_error_jsonld_(diff,error(table_diff_timeout,_), JSON) :-
+    format(string(Msg), "The table diff you requested is too complex for our little brains. Perhaps try signing up for a dedicated table diff account?", []),
```
matko, line 1091: lol. not sure if I'd phrase it that way but it'll do.

src/core/api/api_error.pl
```diff
@@ -1087,6 +1087,13 @@
...
+    JSON = _{'@type' : 'api:DiffErrorResponse',
+             'api:status' : 'api:i_am_a_teapot',
```
matko, line 1093: :D But really, we shouldn't be using this, as terminusdb is in fact not a teapot. Considering this message is trying to push people onto a paid tier, maybe 402 payment required? Apparently there's precedent - wikipedia says google uses it when people exceed limits on their APIs.

matko (changes requested):
As much as I love teapots, I really do want another status code here. Please.

src/core/api/api_error.pl
GavinMendelGleason: payment required it is!

matko (approved):

PR #956, opened 2022-02-02
Add more tests and fix capabilities bug

rrooij:
Some were already covered and others have been added to make some routes.pl tests unnecessary.

I also fixed a bug with regards to authentication. It would fail certain endpoints when tested with header authentication.


rrooij:
Wow, a lot more autoloading issues came up since I started working on this. I will rebase from main and try to resolve the remaining ones. :smile:

spl:
Am I correct in understanding that this is pending on #962?

rrooij:
I think it is ready for review now. I want to convert more tests but I do think that it is worth merging this already in advance because there are important bug fixes like the `capabilities_bug` as well.

spl:
> I've got no clue how these other commits ended up here? I rebased the branch from main and it apparently caused some problems.

I don't know either. I added a commit, rebased to `main`, and force-pushed. Now the changes to `src/core/document/diff.pl` don't show up.

spl:
I'll look at this in more detail tomorrow.

tests/test/triples.js
```diff
@@ -0,0 +1,65 @@
...
+    @prefix owl: <http://www.w3.org/2002/07/owl#> .
+    layer:LayerIdRestriction a owl:Restriction.`
+    const secondReq = await agent.put(path).send({
+      commit_info: { author: 'TestSuite', message: 'Testing Turtle triples 2' },
+      turtle: secondTurtle,
+    })
+    expect(secondReq.status).to.equal(200)
+  })
+
+  // TODO: Create test for this, it responds with an application/json type now... which it isn't.
```
spl, line 61: Please create issues for these TODOs and reference them in the comments here.

tests/lib/agent.js
```diff
@@ -29,16 +29,20 @@ class Agent {
   // Add authentication
   auth () {
     this.userName = process.env.TERMINUSDB_USER
-    assert(this.userName, 'Missing environment variable: TERMINUSDB_USER')
```
spl, line 32: Why do you remove this?

tests/lib/agent.js
```diff
@@ -29,16 +29,20 @@ class Agent {
...
         request.auth(token, { type: 'bearer' })
       })
+    } else if (insecureUserHeader) {
+      this.agent.use((request) => {
+        request.set(insecureUserHeader, this.userName)
+      })
     } else {
       const pass = process.env.TERMINUSDB_PASS
-      assert(pass, 'Missing environment variable: TERMINUSDB_ACCESS_TOKEN or TERMINUSDB_PASS')
+      assert(pass, 'Missing environment variable: TERMINUSDB_ACCESS_TOKEN, TERMINUSDB_INSECRE_USER_HEADER or TERMINUSDB_PASS')
```
spl, line 45:
```suggestion
      assert(pass, 'Missing environment variable: TERMINUSDB_ACCESS_TOKEN, TERMINUSDB_INSECURE_USER_HEADER, or TERMINUSDB_PASS')
```

spl (changes requested):
I'm still reviewing. Here a few comments.

tests/lib/agent.js
rrooij: Because the JWT auth does not need a user, since the user is already encoded in the token.

spl: But that's only for JWT. `this.userName` is used twice below. What happens when it's `undefined` in those cases?

```js
      this.agent.use((request) => {
        request.set(insecureUserHeader, this.userName)
      })
```

```js
      this.agent.use((request) => {
        request.auth(this.userName, pass)
      })
```

rrooij: You're right. It should conditionally assert it instead. :smile:

rrooij: I think I fixed it in the latest commit.

spl (approved):

PR #982, opened 2022-02-10
A new system for id generation

matko:
This is old work for id generation. We want to continue this but had no visibility on it. Therefore, it is now a draft pr.


hoijnet:
Would these be connected to also getting the id issues in subdocuments fixed, so that all is in the same place?

dfrnt-SigmundFuchs:

Issue #720, opened 2021-10-20
Cache important layers


matko:
There are some layers that are opened so often that it really makes no sense to ever forget about them. These are
- the metadata schema graph
- the commit schema graph
- the system schema graph

These three graphs never change during the life of a server. It makes no sense to keep looking them up from a label file and re-opening them.

Additionally, we have the system instance graph. This does change, but not as often as it is read. It makes sense to keep the last known head layer always in memory, and just check if this is still the most recent layer.

In some environments, such as TerminusX, there may be additional databases which are used so often that we'd like to pin them in memory like we would with the system instance graph.

I propose we implement this in terminusdb and terminus-store in the following way:
- The store library gets a new predicate, `head_id/2` which takes a named graph, and returns a layer id for its current head.
- We add a new predicate, `pinned_graph_descriptor(+Graph_Name)` which yields true for graph descriptors which we wish to pin in memory. We can hardcode the system instance graph, and maybe provide additional ways to pin particular databases through environment variables.
- We add a new predicate, `permanent_pinned_graph_descriptor(+Graph_Name)` which yields true for graph names which we wish to pin in memory and which we know will never change (meaning, we never need to recheck the label file / meta graph).
- We add a dynamic predicate, `cached_layer(+Graph_Name, -Layer)`, which can be used to cache a layer for a particular graph name.
- `descriptor:open_read_write_obj` clauses check the given graph descriptor against `pinned_graph_descriptor/1` and `permanent_pinned_graph_descriptor/1`. If it matches, it uses `cached_layer/2` to retrieve the cached version, and checks with `head_id/2` if this is still the valid head layer (for non-permanent pinned graphs). If not, the layer is opened normally and asserted on `cached_layer/2`.

This should ensure that we have to read from disk far less often, and for common unchanging graphs, that we don't have to reread from disk at all. It also requires only a minimal change in the store library. Most of the behavior is implemented in prolog.

matko:
New approach is to do it all in prolog. There's no need for any special caching to happen from rust, as keeping around the atom referring to a layer in prolog is enough to keep it around. This allows us to have all pinning logic in prolog.

PR #1004, opened 2022-02-17
Implement layer pinning

matko:
# layer pinning
## schema graphs for metadata
Most easy, always the same layer, users can't change it, can cache the initial result and never throw it away

## system database
This can change, so we have to check that what we're opening is still the most recent one.

## Arbitrary user databases
This is the most difficult, because we need to think of policy around when layers should be evicted from memory.

### DIFF
--- a/.lint_config.pl
+++ b/.lint_config.pl
@@ -2,3 +2,4 @@
 ignore_predicate("dateTimeStamp/11").
 ignore_predicate("time/8").
 ignore_predicate("dateTime/11").
+ignore_predicate("woql_compile/5").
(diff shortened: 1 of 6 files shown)


matko:
There is a slight approach with the current approach to pinning.
It stores a list of layers along with a descriptor, and whenever there might be cause to refresh this cached version, we compare this layer list with the newly retrieved or generated layer list from open_descriptor or from a transaction commit.
This comparison is done by comparing the layer ids. This is however not sufficient. Two layers may be different even if their id matches. This happens when we optimize layers - both the optimized (rolled up) version and the original layer will share a layer id, but we definitely want to keep the optimized verison around, not the unoptimized! Unfortunately, the way things are set up now, we'll be keeping the unoptimized version around instead.

What we need to do is build a better comparer in the rust library which doesn't compare the layer id, but the actual underlying pointer to the layer data structure. This should actually also be a much cheaper check than the existing compare based on layer id.

matko:
This implements #720.

GavinMendelGleason (approved):

PR #1005, opened 2022-02-17
Add missing exports to fix unit tests

matko:
Two predicates were missing from terminus_config.pl, which were causing unit tests to fail locally (though not in the docker container and therefore not on CI).

These are now exported properly.

### DIFF
--- a/src/config/terminus_config.pl
+++ b/src/config/terminus_config.pl
@@ -25,7 +25,9 @@
               insecure_user_header_key/1,
               check_all_env_vars/0,
               is_enterprise/0,
-              check_insecure_user_header_enabled/1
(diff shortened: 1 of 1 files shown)


spl:
Does this need changes to pass `lint`?

matko:
> Does this need changes to pass `lint`?

Apparently. I'll look into that today.

matko:
It complains about `skip_generate_nsols/3` not being found in two places. The strange thing is that in the two places it mentions, that is not even called, and even if it were called, it's being imported from util. I have no idea what is going on. @rrooij ?

spl (approved):
We will look into the lint errors later.

PR #1006, opened 2022-02-18
Fix QueryResource post and url, add tests

spl:
* `copy_remote`: catch `http_open` errors, wrap them, and re-throw them
* `QueryResource`: throw error for missing file
* Add tests originally by @Neelterminusdb
* Add `http-server` for testing with served files
* Add `tests/tests-served` for these tests
* Run Docker in CI with `--net=host` to allow it to work with
  `http-server`


GavinMendelGleason (approved):
Looks good!

PR #1012, opened 2022-02-22
Remove jwt support

rrooij:
We rely on a Prolog JWT library which is not very well maintained anymore. API gateways like Kong provide
better JWT support and should be used instead of doing the same verification twice.

The insecure user header can be used instead to configure a proxy to check for the JWTs.

The Python tests fail because they still test for JWT. If we decide to merge this, we should remove the JWT test from the Python client.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,8 +150,6 @@ jobs:
             --net=host \
             -e TERMINUSDB_INSECURE_USER_HEADER_ENABLED=true \
             -e TERMINUSDB_INSECURE_USER_HEADER='X-Forwarded-User' \
-            -e TERMINUSDB_JWT_ENABLED=true \
-            -e TERMINUSDB_SERVER_JWKS_ENDPOINT='https://cdn.terminusdb.com/jwks.json' \
             $DOCKER_IMAGE_NAME:local
 
       - name: Set up Node.js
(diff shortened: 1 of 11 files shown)


hoijnet:
I saw the direction of removing the JWT support, and tend to disagree with the insecure headers being used instead. A failure of networking provisions would leave the instance open to attacks, as would middleware configuration errors that would leave options for attackers to impersonate users easily by just guessing usernames. Latency-wise, the extra validation step is very short, and I would propose that a layered approach with multiple JWT inspections reduce the risk of a single implementation failing.

Relying on the insecure user header is, as the insecure word implies, insecure. Please reconsider, I feel the JWT support is important and I feel it is important for the platform.

rrooij:
Thanks @hoijnet for sharing your concerns. We are not planning to drop it right away but decided to mark it as 'deprecated' until a solution to the unmaintained Prolog library shows up. Perhaps we'll remove it in the future if no maintainer shows up, but not without proper communication on how to properly set up TerminusDB in such a way without JWT in TerminusDB itself.

I do disagree with your security concerns as in the proper set-up, TerminusDB will be entirely closed of to the outside world. I agree that it is not Zero Trust architecture. But we also have to keep in mind that it can also provide a false sense of security if an API gateway isn't used, since the Prolog JWT library is not well maintained and is therefore is a bit risky to rely on it.

hoijnet:
Thanks, got a bit worried, and share your concerns regarding the unmaintained library! 

A thought briefly crossed my mind in relation to the web service layer of TerminusDB being in Prolog rather than in Rust which is getting faster and probably better for serving web; with Rust probably the stronger language for the web server layer for many reasons, including having more eyes on the implementations of the various parts. However it's a tall order to refactor, and not worth it for just the JWT at the moment.
 
It would probably be hard to get the header set in WOQL-sourced connections out from the server, but I'm thinking that any exposed ports should be authenticated for such reasons. Especially when connections can be made on the clients behalf out of TerminusDB with WOQL.

I understanding the reasoning better now, thanks! Prioritizing up to date library dependencies is comforting; and also a hard (but good) choice to make! As always, tradeoffs are hard.

Thanks!!

hoijnet:
And, I should add, thanks for making the decision of making the change through deprecation!

PR #1013, opened 2022-02-24
replace_document: check submitted ID against generated

spl:
* Avoids a problem with the random key type where the submitted document does not contain an `@id`.
* Consolidate `document_does_not_exist` test into `document_not_found`
* Test for `document_not_found`
* Test for WOQL `UpdateDocument` with various configurations of IDs
* Refactor `json_idgen`: Lift out `key_descriptor`, pattern match on it in `json_idgen`

### DIFF
--- a/src/core/api/api_error.pl
+++ b/src/core/api/api_error.pl
@@ -166,6 +166,35 @@ api_global_error_jsonld(error(http_open_error(Err), _), Type, JSON) :-
                               'api:reason' : Err },
              'api:message' : Msg
             }.
+api_global_error_jsonld(error(document_not_found(Id), _), Type, JSON) :-
+    error_type(Type, Type_Displayed),
+    format(string(Msg), "Document not found: ~q", [Id]),
+    JSON = _{'@type' : Type_Displayed,
+             'api:status' : "api:not_found",
+             'api:error' : _{ '@type' : 'api:DocumentNotFound',
+                              'api:document_id' : Id },
+             'api:message' : Msg
+            }.
+api_global_error_jsonld(error(document_not_found(Id, Document), _), Type, JSON) :-
+    error_type(Type, Type_Displayed),
+    format(string(Msg), "Document not found: ~q", [Id]),
+    JSON = _{'@type' : Type_Displayed,
+             'api:status' : "api:not_found",
+             'api:error' : _{ '@type' : 'api:DocumentNotFound',
+                              'api:document_id' : Id,
+                              'api:document': Document },
+             'api:message' : Msg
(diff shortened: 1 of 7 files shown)


src/core/api/api_error.pl
```diff
@@ -166,6 +166,35 @@
...
+api_global_error_jsonld(error(document_not_found(Id), _), Type, JSON) :-
+    error_type(Type, Type_Displayed),
+    format(string(Msg), "Document not found: ~q", [Id]),
+    JSON = _{'@type' : Type_Displayed,
+             'api:status' : "api:not_found",
+             'api:error' : _{ '@type' : 'api:DocumentNotFound',
+                              'api:document_id' : Id },
+             'api:message' : Msg
+            }.
+api_global_error_jsonld(error(document_not_found(Id, Document), _), Type, JSON) :-
```
matko, line 178: Do we actually generate this in a way that it appears at this level of error transform? document api endpoints should be transforming this to include the document.

matko: This is what `call_catch_document_mutation` does, it checks for `known_document_error` and if the case is included, it makes sure to insert the document it was currently inserting in there. Is that appropriate in this case?

spl: Yeah, this confused me for a bit. So, `document_not_found` does not actually appear in `known_document_error`.

matko (approved):

src/core/document/json.pl
```diff
@@ -368,39 +368,54 @@
...
+    get_field_values(JSON, DB, Context, Fields, Values),
+    path_component([type(Base)|Path], Context, [Path_Base]),
+    idgen_hash(Path_Base, Values, Id).
+json_idgen(value_hash(Base), JSON, _DB, _Context, _Path, Id) :-
+    !,
+    get_all_path_values(JSON, Path_Values),
+    idgen_path_values_hash(Base, Path_Values, Id).
+json_idgen(Descriptor, JSON, _DB, Context, Path, Id) :-
+    (   Descriptor = random(Base)
+    ;   Descriptor = base(Base)
```
GavinMendelGleason, line 387: This is better than before! However, stylistically, it's slightly better to have no explicit cuts, and to instead use first argument indexing, which means no fall-through case (two clauses for `random` and `base`).

spl: Is it better now?

GavinMendelGleason (approved):

src/core/document/json.pl
GavinMendelGleason: Yup!

PR #1018, opened 2022-03-10
Log verbose output to stderr and print proper JSON

rrooij:
This makes it easier to pipe the JSON to a file using stdout for instance by using `node bench.js > results.json`.

Also, the objects were not proper JSONs but JavaScript objects


rrooij:
Better?

PR #1025, opened 2022-03-15
Added version number to startup banner

matko:
For the sake of better logs, we should always print the version number at startup, so that when someone submits logs to us as part of their error report, we know what version that bug occurred on.


rrooij:
`ERROR: ./src/server/main.pl:114| Predicate "terminusdb_version/1" not found` in linting

rrooij (approved):

PR #1035, opened 2022-03-17
$TERMINUSDB_DOCKER_IMAGE_TAG: ci.yml, terminusdb.sh

spl:
* `ci.yml`:
  - Use `$TERMINUSDB_DOCKER_IMAGE_TAG` for the Docker image tag name
* `tests/terminusdb.sh`:
  - Use environment variables including `$TERMINUSDB_DOCKER_IMAGE_TAG`
  - Better error reporting

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,10 @@ on:
   workflow_dispatch:
 
 env:
+  TERMINUSDB_DOCKER_IMAGE_NAME: terminusdb/terminusdb-server
+  TERMINUSDB_DOCKER_IMAGE_TAG: terminusdb/terminusdb-server:local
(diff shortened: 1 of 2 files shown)


rrooij:
Makes a lot of sense, although I would probably split the name of the image and the tag name. The tag name also contains the image name now and I think it would be better to construct it as a new variable with terminusdb_name and terminusdb_tag appended to each other.

spl:
@rrooij I wanted to do that in the `env`, but it doesn't allow you to have variables refer to each other. What's the best way to do that with the least redundancy?

rrooij:
Hmmm I found this: https://brandur.org/fragments/github-actions-env-vars-in-env-vars

But that does not seem like a clean solution either. Lets just stick with this. :smile:

rrooij (approved):

PR #1095, opened 2022-04-08
Unify full_replace and non-full_replace insert

### LINKED ISSUE #1063: Full replace insert does not get checked for duplicate IDs as normal insert
**Describe the bug**
On adding an Enum to an already existing schema - fails
```
@type: "not_a_class_or_base_type"
class: "iri://CAMS/Cons/d890d18e2e838884e1c9909214914f20a7c0c998302d13e097b10831bbc948b5"
```
**To Reproduce**
1. Add an enum with an existing document name - 
 ```
@type: "not_a_class_or_base_type"
class: "iri://CAMS/Cons/d890d18e2e838884e1c9909214914f20a7c0c998302d13e097b10831bbc948b5"
```

**Expected behavior**
Better error message - saying to change the name of Enum and make it different from a document name

### PR DESCRIPTION
**Pull request by spl**
* `full_replace` schema insert was using a different code path, which
  lead to a strange error when the same `@id` is used in multiple types.
* This change makes instance and schema insert use the same code paths
  for both `full_replace` and not.
* Rename some `json_read` predicates for clarity.
* Add `json_read_list_stream_head` and `json_read_tail_stream` for
  reading the first term from a stream and then the rest.
* Add unit tests to `json_read` predicates.
* Fixes terminusdb/terminusdb#1063

### CODE REVIEW

File: src/core/api/api_document.pl [hunk 1]
```diff
@@ -165,14 +164,13 @@
...
-replace_existing_graph(schema, Transaction, Stream, _Captures_Var) :-
-    replace_json_schema(Transaction, Stream).
-replace_existing_graph(instance, Transaction, Stream, Captures_Var) :-
-    [RWO] = (Transaction.instance_objects),
-    delete_all(RWO),
-    forall(json_read_dict_list_stream(Stream, Document),
-           nb_thread_var({Transaction,Document}/[State,Captures_Out]>>(api_insert_document_(instance, Transaction, Document, State, _, Captures_Out)),
-                         Captures_Var)).
+prepare_full_replace(true, schema, Stream, Transaction, Tail_Stream) :-
+    % Read a new context from the stream and replace the existing one.
```
matko, line 168: should this not cut?

src/core/api/api_document.pl
```diff
@@ -183,18 +181,24 @@
...
-                                     Ids),
-
-                             die_if(nonground_captures(Captures_Var, Nonground),
-                                    error(not_all_captures_found(Nonground), _)),
-                             die_if(has_duplicates(Ids, Duplicates), error(same_ids_in_one_transaction(Duplicates), _)))),
+                         ->  api_nuke_documents_(Graph_Type, Transaction)
+                         ;   true
+                         ),
+                         prepare_full_replace(Full_Replace, Graph_Type, Stream, Transaction, Tail_Stream),
+                         ensure_transaction_has_builder(Graph_Type, Transaction),
```
matko, line 188: you should ensure the builder before you prepare the full replace as that already potentially uses the builder.

matko (changes requested):
I like the approach to unify stream and list reading with the head/tail stuff.

spl:
Okay, this is ready to go now. We figured out that the failing “Check / Python client tests is due to <https://github.com/terminusdb/terminusdb/issues/1099>.

matko (approved):

Issue #1062, opened 2022-03-28
Occasional benchmark failure


spl:
At least one of the benchmarks is failing on a weekly basis during CI runs. Below is the output from one.

It's not always the same benchmark. It seems to usually be a document insert with 1,000 documents, not smaller. I've seen it happen with both schema and instance inserts.

```
>>> bench/document-instance-insert-first-1000-simple.js (>=5 samples)
AssertionError: expected 400 to equal 200
    at Object.verifyInsertSuccess (/home/runner/work/terminusdb/terminusdb/tests/lib/document.js:183:23)
    at h.afterEach (/home/runner/work/terminusdb/terminusdb/tests/bench/document-instance-insert-first.template.js:25:16)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async o (/home/runner/work/terminusdb/terminusdb/tests/node_modules/buffalo-bench/dist/index.min.js:1:942)
    at async h.runSample (/home/runner/work/terminusdb/terminusdb/tests/node_modules/buffalo-bench/dist/index.min.js:1:2830)
    at async h.run (/home/runner/work/terminusdb/terminusdb/tests/node_modules/buffalo-bench/dist/index.min.js:1:3109)
    at async main (/home/runner/work/terminusdb/terminusdb/tests/bench.js:66:5) {
  code: undefined,
  name: 'AfterEachError',
  statusCode: 2,
  showDiff: true,
  actual: 400,
  expected: 200,
  operator: 'strictEqual'
}
```

spl:
I'll add something to print the response for the benchmarks.

spl:
We haven't seen this in a few weeks. Closing as unreproducible.

PR #1101, opened 2022-04-12
set the document handler to chunked to support streaming output

matko:
This adds `chunked` to the configuration of the document handler. Doing this should allow streaming output, though unfortunately I don't have a local test set large enough to actually test this properly.

### DIFF
--- a/src/server/routes.pl
+++ b/src/server/routes.pl
@@ -421,6 +421,7 @@ test(triples_update, [
 :- http_handler(api(document/Path), cors_handler(Method, document_handler(Path), [add_payload(false)]),
                 [method(Method),
                  prefix,
+                 chunked,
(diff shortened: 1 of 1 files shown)


matko:
Testing it a bit more, it's definitely sending it chunked in chunks of 1000 characters. So this should stream well.

spl:
How do the JavaScript and Python clients handle it?

matko:
looks like a bunch of tests are failing vaguely so maybe they're not handling it so well.

matko:
It seems to work just fine for curl though.

spl:
Those failures are known:
* https://github.com/terminusdb/terminusdb/issues/1062
* https://github.com/terminusdb/terminusdb-client-python/issues/298

PR #1114, opened 2022-04-21
Use 'unsafe' document insert for full_replace

### LINKED ISSUE #1112: Can't insert same document with full_replace twice
PR #1095 broke document insert with `full_replace`. If you try to insert a document with the same `@id` as an existing document, it will fail with an error like:

```js
{
  '@type': 'api:InsertDocumentErrorResponse',
  'api:error': {
    '@type': 'api:DocumentIdAlreadyExists',
    'api:document': { '@id': 'eb15db', '@type': 'Class' },
    'api:document_id': 'eb15db'
  },
  'api:message': 'Tried to insert a new document with id "eb15db", but an object with that id already exists',
  'api:status': 'api:failure'
}
```

spl:
After #1095, we started using the 'normal' document insert for
full replace, but it checked for the existence of the document ID on
insert, which meant it failed when inserting the same document again.

Now, we use an 'unsafe' document insert for full replace, which doesn't
check for the ID. We already nuke all documents, so this is fine.

This also adds integration tests which were missing from #1095.

Fixes #1112


src/core/api/api_document.pl
```diff
@@ -164,14 +163,53 @@
...
+        error(document_insertion_failed_unexpectedly(Document), _)),
+    do_or_die(
+        Id = (Document.get('@id')),
+        error(document_has_no_id_somehow, _)).
+api_insert_document_unsafe_(instance, Transaction, Document, state(Captures_In), Id, Captures_Out) :-
+    do_or_die(
+        insert_document_unsafe(Transaction, Document, Captures_In, Id, Captures_Out),
+        error(document_insertion_failed_unexpectedly(Document), _)).
+
+insert_documents_(true, Graph_Type, Stream, Transaction, Captures_Var, Ids) :-
```
GavinMendelGleason, line 178: This looks good. Note though that it might be slightly better to use first-argument indexing with a discriminator (in this case true/false) than testing a first argument, and using a cut.

GavinMendelGleason (approved):

src/core/api/api_document.pl
spl: https://github.com/terminusdb/terminusdb/pull/1117

PR #1115, opened 2022-04-21
Use random storage subdirectory for CLI tests

### LINKED ISSUE #1085: cli integration tests reinitializes store rather than create a test one, then fail
It looks like the cli integration tests intend to create a storage/ folder in the tests/ directory using the terminusdb.sh script. However, when run locally, against a locally built terminusdb binary, this actually does not create a storage/ folder inside of tests/, but reinitializes the storage directory one directory up, then fail the test run because afterwards it is unable to delete tests/storage.

Reinitializing local store is bad!

spl:
Currently, the CLI integration tests force initialize the store, which
removes any existing default store (`./storage/db`). This may not be
desirable if you are currently using that store for development.

However, sometimes, the tests fail or are killed, and it is useful to
remove any existing store when re-running the tests.

This change uses the `TERMINUSDB_SERVER_DB_PATH` environment variable to
set the store to a random subdirectory of `./storage`. We can therefore
both avoid removing the default store and remove any previously created
random subdirectory used for tests.

Fixes #1085.


spl:
I forgot to handle the CLI tests running in the Docker container. But this is now ready for review.

GavinMendelGleason (approved):

PR #1120, opened 2022-04-22
Initial basic document insert and get in CLI

spl:
Add `doc insert` and `doc get` subcommands to the TerminusDB CLI. Initially, these will allow us to test the CLI more easily. The input is stdin, and the output is stdout. The options available are limited. In the future, we will add more functionality to these subcommands.


GavinMendelGleason (approved):
Really well written

PR #1163, opened 2022-05-09
ci: don't delete old docker images on fork

rrooij:
They don't have access to the secrets which always make the CI cron fail.


spl (approved):
Exactly. Thanks!

Issue #1020, opened 2022-03-14
Test push and pull


spl:
* Test if the `push` and `pull` operations work against a TerminusDB database.
* Make sure it works with the TerminusX test database.

spl:
* [ ] Add API tests for `push`, `pull`, and `clone`

hoijnet:
This has been checked by DFRNT to work. Care has to be taken to align on SSL interfaces to be compatible with SWIPL and tus but works as they should and has passed our tests for them. Closing.

PR #1165, opened 2022-05-11
CLI clone, push, pull improvements & tests

### LINKED ISSUE #1127: Clone failure leads to inconsistent database
A database created as a result of a clone is in an inconsistent state if there is a failure - this should have a cleanup attached which removes the database - this should probably use `setup_call_catcher_cleanup/4` directly *after* database construction, or filtering out the database exists exception, so as not to destroy already existing databases. The removal has to be *force* deleted as it is a partial DB construction finalised by clone.

### LINKED ISSUE #1164: tmp_file_stream/3 does not produce unique file names over multiple Docker runs for push
This library uses [`tmp_file_stream/3`](https://www.swi-prolog.org/pldoc/doc_for?object=tmp_file_stream/3) to create a temporary file. This file name is used to create a hash for the remote file name. So, if the same file name is created multiple times in a row, the later uploads will fail.

The TerminusDB CI tests use a Docker container, and `tmp_file_stream/3` produced the same file name for every `push` command. It seems (in [`TemporaryFile()`](https://github.com/SWI-Prolog/swipl/blob/7fb76e51c4d2954031df49ddc9bb335bf9cfd808/src/os/pl-os.c#L706-L796)) to use the process ID (`getpid()`) and a temporary counter (`temp_counter`) for uniqueness. I was seeing the file path `/tmp/swipl_1_22` repeatedly, so I presume the PID is 1 and the counter is 22 on every `docker run` of the image.

(Aside: `TemporaryFile()` is also used by [`tmp_file/2`](https://www.swi-prolog.org/pldoc/doc_for?object=tmp_file/2).)

I thought we might perhaps instead use the `mktemp` command, which is available on Linux and macOS, and does produce unique file names across different container runs. It's interesting to note that the [`mktemp/2`](https://www.swi-prolog.org/pldoc/doc_for?object=sicstus_system%3Amktemp/2) predicate (a) is documented as deprecated because it is a security risk and (b) uses `tmp_file/2` [under the hood](https://github.com/SWI-Prolog/swipl/blob/3078b30b9fa69ca4ae3f36dcd9304efefd1d04b4/library/dialect/sicstus/system.pl#L241-L242).

A quick solution would be for our TerminusDB Docker usage in testing to mount the host system `/tmp` directory to the container `/tmp`, which would avoid the problem in the immediate timeframe. But I can imagine this problem affecting others in the future, and it is a difficult one to track down.

spl:
* Improve stdout messages
* Improve exception catching and error messages
* Handle connection and auth failures in clone
* Change `most_recent_common_ancestor/7` to handle initial commits
* Part of #1020
* Fixes #1127
* Fixes #1164


src/cli/main.pl
```diff
@@ -980,50 +986,53 @@
 
     create_authorization(Opts,Authorization),
 
+    format(current_output, "Cloning: origin~n", []),
```
matko, line 989: the message formatting suggests this could say something else than 'origin' but that's not the case. Maybe this should give some more detail like 'cloning from <remote> into new database <db>'?

src/cli/main.pl
```diff
@@ -980,50 +986,53 @@
...
+    ;   Pull_Status = "api:pull_fast_forwarded"
+    ->  format(current_output, "Fast-forward pull.", [])
+    ;   format(current_output, "Pull status: ~w.", [Pull_Status])
+    ),
+    (   Head_Updated = true
+    ->  format(current_output, " Branch head updated.", [])
+    ;   Head_Updated = false
+    ->  format(current_output, " Branch head not updated.", [])
+    ;   format(current_output, " Branch head update status: ~w.", [Head_Updated])
+    ).
```
matko, line 1024: Some comments on these messages:
- You're misinterpreting pull_status and fetch_status. fetch_status (which you use for Head_Updated) indicates whether or not we pulled in any changes from the remote. It does not indicate a local branch change. It is pull_status which indicates whether or not a branch changed, and if it didn't change, why it didn't change. It is possible for `fetch_status` to be `true` and `pull_status` to be `pull_unchanged` if none of the remote changes impacted the branch we're pulling into. Likewise, it is possible for `fetch_status` to be `false` and `pull_status` to be something other than `pull_unchanged` if our fetched copy of the repository was already up to date but not yet integrated in our local repository (such as when the user fetches first, or when they previously pulled some other branch triggering that fetch).
- whenever we say we didn't do something, I think it's a good idea to say why we didn't, to prevent users from getting frustrated about something that they feel should have happened but didn't.
- imo messages should include information about what branches and what remotes they operate on.

src/core/api/db_push.pl
```diff
@@ -128,15 +128,7 @@
...
-        catch(call(Push_Predicate, Remote_URL, Payload),
-              E,
-              (   E = error(Inner_E,_),
-                  (   Inner_E = history_diverged
-                  ;   Inner_E = remote_unknown
-                  ;   Inner_E = authorization_failure(_)
-                  ;   Inner_E = communication_failure(_))
-              ->  throw(error(remote_unpack_failed(Inner_E),_))
-              ;   throw(error(remote_unpack_unexpected_failure(E),_)))),
+        call(Push_Predicate, Remote_URL, Payload),
```
matko, line 131: how come we can do away with those catches?

matko (changes requested):
This looks mostly good. I like the part where you changed the recent common ancestor calculation to do something meaningful for when there's no common ancestry. My main concern is with the messages around push which I currently find misleading. I also wonder if/how your changes affected clone/push/pull from the api rather than from the cli.

src/cli/main.pl
spl: The remote name `origin` is hardcoded for the `clone` operation but not for other operations. I added it here, because you otherwise do not know which remote name is being used. (I think we eventually should add a flag for the remote name to be consistent with other commands. I've created #1167 for that.).

It does say `Database created` after the operation is completed below. I used `Database created` to be consistent with the message of `db create`.

The URL cloned from is in the command argument, so I elided that to avoid redundancy. But if you think we should put the URL here, I'm fine with that.

Would it be better for you if the message was `Cloning from remote 'origin'`?

src/cli/main.pl
spl: I used  the variable name `Head_Updated` because `src/core/api/db_pull.pl` used `Head_Has_Updated` as a variable for `api:fetch_status`. With the name `api:fetch_status`, I have no idea what the value `true` or `false` means. I believe my mistake is in not realizing it was the _remote_ branch head that was updated, not the _local_ branch head.

I would be happy to provide more information in the messages. Perhaps we need to talk about how to phrase these, because providing the JSON in the CLI messages is not enough.

matko: For `Head_Updated`, it's actually neither the remote nor local branch head that was updated. the `Head` in `Head_Updated` refers to the head of the commit graph for the remote repository. It could be that this was updated due to remote changes, but none of these changes are in the branch we're pulling from. So if `true`, the commit graph for the remote repository has a new head (which we just fetched). If `false`, the commit graph for the remote repository was up to date, and nothing was actually downloaded from the remote.

src/core/api/db_push.pl
spl: Because we catch the appropriate errors in `api_report_errors`. I don't see why we need to wrap these in `remote_unpack_failed` and `remote_unpack_unexpected_failure`. Thanks to the first argument `push` to `api_report_errors`, we know where the error is coming from. Also, `authorized_push` was simplified to wrap connection errors with `remote_connection_failure`, and `api_global_error_jsonld` now attempts to print nice error messages for `remote_connection_failure`.

src/cli/main.pl
spl: Ah, okay. Thanks for the explanation about the commit graph. So, for `api:fetch_status`, we could simply report if the remote commit graph was updated?

src/cli/main.pl
matko: > Would it be better for you if the message was `Cloning from remote 'origin'`?
That'd be better. Even better imo would be to include the URI that is being cloned, but that's nitpicking.

spl: > Even better imo would be to include the URI that is being cloned, but that's nitpicking.

Do you mean the remote URI that is passed to the `clone` command? As I said, that's just a bit redundant, because it's already there, but I'm fine with it. If we do that with `clone`, then I would also change the other commands to reprint some of the arguments for consistency.

Or do you mean the actual API URI used internally? If so, that might involve changing the `clone` predicate in `db_clone.pl`

PR #1174, opened 2022-05-12
Payload enterprise sha

rrooij:
Instead of always cloning main, the payload SHA should be passed.


.github/workflows/ci.yml
```diff
@@ -153,10 +153,11 @@ jobs:
       github.event_name == 'push' &&
       github.ref == 'refs/heads/main'
 
+
```
spl, line 156:
```suggestion
```

spl (approved):

PR #1178, opened 2022-05-13
Accept test_ref in check.yml

spl:
This allows us to use `check.yml` outside of the repository and
guarantee that every use of `actions/checkout` gets the same ref. It has
happened that the `HEAD` of the default branch changes in the middle of
these multiple checkouts. With this change, that won't happen.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,7 @@ jobs:
     uses: ./.github/workflows/check.yml
     with:
       test_repository: ${{ github.repository }}
+      test_ref: ${{ github.ref }}
 
   benchmark:
     name: Benchmark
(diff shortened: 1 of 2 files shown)


rrooij (approved):

.github/workflows/ci.yml
```diff
@@ -71,6 +71,7 @@ jobs:
     uses: ./.github/workflows/check.yml
     with:
       test_repository: ${{ github.repository }}
+      test_ref: ${{ github.ref }}
```
rrooij, line 74: WAIT. I merged prematurely. I think this should be the commit hash instead of the ref. So this should be `${{ github.sa }}

spl: https://github.com/terminusdb/terminusdb/pull/1179

PR #1191, opened 2022-05-17
Generic JSON insertion and retrieval

GavinMendelGleason:


src/core/document/json_rdf.pl
```diff
@@ -0,0 +1,505 @@
...
+    R = X^^T.
+json_type_rdf_type(X,R),
+atom(X),
+X = null =>
+    global_prefix_expand(xsd:token, T),
+    R = X^^T.
+
+json_object_triple(Parent_Id, Parent_Prop, Object, Triple),
+is_dict(Object) =>
+    variant_sha1(Object, Hash),
```
GavinMendelGleason, line 53: Should use crypto - sha2

src/core/document/json.pl
```diff
@@ -10849,3 +10934,334 @@
...
+            )),
+      cleanup(teardown_temp_store(State)),
+      error(
+          not_a_valid_json_object_id('terminusdb:///data/named'),
+          _)
+     ]) :-
+
+    Document =
+    json{
+        '@id' : named,
```
GavinMendelGleason, line 11214: How can this be named `named`

src/core/document/json.pl
```diff
@@ -10849,3 +10934,334 @@
...
+
+    with_test_transaction(
+        Desc,
+        C1,
+        insert_document(C1,Document,true,Id)
+    ),
+
+    Document2 =
+    json{
+        '@id' : 'JSON/my_json',
```
GavinMendelGleason, line 11114: If top level objects are `JSONDocument` this should be `JSONDocument/`

src/core/document/json.pl
```diff
@@ -2633,16 +2691,37 @@
...
-replace_document(Transaction, Document, Create, Captures_In, Id, Dependencies, Captures_Out) :-
+replace_document(Transaction, Document, Create, true, Captures, Id, [], Captures) :-
+    is_transaction(Transaction),
+    !,
+    database_prefixes(Transaction, Prefixes),
+    do_or_die(
+        (   del_dict('@id', Document, Id, JSON)
+        ->  prefix_expand(Id, Prefixes, Id_Ex),
+            (   is_json_hash(Id_Ex)
+            ->  assign_json_object_id(JSON, Id_Ex)
```
GavinMendelGleason, line 2714: This should be random, and not a hash so that replacement is possible.

src/core/api/api_document.pl
```diff
@@ -190,29 +191,43 @@
...
+    option(graph_type(Graph_Type), Options),
+    option(author(Author), Options),
+    option(message(Message), Options),
+    option(full_replace(Full_Replace), Options),
+    option(json(JSON), Options),
+    die_if(
+        (   Graph_Type = schema,
+            JSON = true
+        ),
+        error(json_and_schema_disallowed,_)
```
spl, line 229: This error should be caught, a useful error message should be printed, and a test should be added for that message.

src/cli/main.pl
```diff
@@ -1221,67 +1227,56 @@
...
     api_report_errors(
         replace_documents,
         (   (   var(Data)
-            ->  with_memory_file(doc_replace_memory_file(System_DB, Auth, Path, Graph_Type, Author, Message, Create, Ids))
+            ->  with_memory_file(doc_replace_memory_file(System_DB, Auth, Path, Ids, Opts))
             ;   open_string(Data, Stream),
-                api_replace_documents(System_DB, Auth, Path, Graph_Type, Author, Message, Stream,
-                                      Create, no_data_version, _, Ids)
+                api_replace_documents(System_DB, Auth, Path, Stream,
+                                      no_data_version, _, Ids, Opts)
```
spl, line 1279:
```suggestion
                api_replace_documents(System_DB, Auth, Path, Stream, no_data_version, _, Ids, Opts)
```

src/cli/main.pl
```diff
@@ -1416,36 +1410,36 @@
...
     % Read the memory file to insert documents.
-    with_memory_file_stream(Mem_File, read, doc_replace_stream(System_DB, Auth, Path, Graph_Type, Author, Message, Create, Ids)).
+    with_memory_file_stream(Mem_File, read, doc_replace_stream(System_DB, Auth, Path, Ids, Opts)).
 
-doc_replace_stream(System_DB, Auth, Path, Graph_Type, Author, Message, Create, Ids, Stream) :-
-    api_replace_documents(System_DB, Auth, Path, Graph_Type, Author, Message, Stream,
-                          Create, no_data_version, _, Ids).
+doc_replace_stream(System_DB, Auth, Path, Ids, Opts, Stream) :-
+    api_replace_documents(System_DB, Auth, Path, Stream,
+                          no_data_version, _, Ids, Opts).
```
spl, line 1421:
```suggestion
    api_replace_documents(System_DB, Auth, Path, Stream, no_data_version, _, Ids, Opts).
```

src/cli/main.pl
```diff
@@ -1416,36 +1410,35 @@
...
     % Read the memory file to insert documents.
-    with_memory_file_stream(Mem_File, read, doc_delete_stream(System_DB, Auth, Path, Graph_Type, Author, Message)).
+    with_memory_file_stream(Mem_File, read, doc_delete_stream(System_DB, Auth, Path, Opts)).
 
-doc_delete_stream(System_DB, Auth, Path, Graph_Type, Author, Message, Stream) :-
-    api_delete_documents(System_DB, Auth, Path, Graph_Type, Author, Message, Stream,
-                         no_data_version, _).
+doc_delete_stream(System_DB, Auth, Path, Opts, Stream) :-
+    api_delete_documents(System_DB, Auth, Path, Stream,
+                         no_data_version, _, Opts).
```
spl, line 1430:
```suggestion
    api_delete_documents(System_DB, Auth, Path, Stream, no_data_version, _, Opts).
```

src/core/document/json.pl
```diff
@@ -2563,9 +2589,31 @@
     ).
 
 insert_document(Transaction, Document, ID) :-
+    insert_document(Transaction, Document, false, ID).
+
+insert_document(Transaction, Document, JSON, ID) :-
```
spl, line 2594: Using `JSON` as an argument name here is very confusing. Here, `JSON` is `true` or `false`, but elsewhere, `JSON` is mainly used for a dictionary.

src/cli/main.pl
```diff
@@ -1221,67 +1227,56 @@
...
         ;   open_string(Data, Stream),
-            doc_insert_stream(System_DB, Auth, Path, Graph_Type, Author, Message, Full, Ids, Stream)
+            doc_insert_stream(System_DB, Auth, Path, Ids, Opts, Stream)
         )
     ),
     length(Ids, Number_Inserted),
     (   Number_Inserted = 1
-    ->  format(current_output, "Document inserted~n", [])
-    ;   format(current_output, "Documents inserted: ~d~n", [Number_Inserted])
+    ->  format(current_output, "Document inserted ~q~n", [Ids])
```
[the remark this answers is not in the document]
GavinMendelGleason: I'd actually prefer the reverse, where you get a `--quiet` but it is in fact really important in practice because Ids can be randomly generated after which point you haven't a clue what you've added.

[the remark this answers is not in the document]
GavinMendelGleason: We can add `--quiet` in a subsequent PR.

src/server/routes.pl
```diff
@@ -485,10 +485,17 @@
             param_value_search_message(Search, Message),
             param_value_search_graph_type(Search, Graph_Type),
             param_value_search_optional(Search, full_replace, boolean, false, Full_Replace),
+            param_value_search_optional(Search, json, boolean, false, JSON),
```
spl, line 488: It was only after reading `insert_document` that I realized this `JSON` that was being passed around was not a dictionary but a boolean. Can we use a different name here? `Is_JSON` or something else?

src/core/document/json.pl
```diff
@@ -2912,10 +2995,12 @@
...
     check_json_string('@id', Id),
+    database_prefixes(Transaction, Prefixes),
     database_schema(Transaction, Schema),
-
-    database_prefixes(Transaction, Context),
-    prefix_expand_schema(Id,Context,Id_Ex),
+    prefix_expand_schema(Id,Prefixes,Id_Ex),
+    do_or_die(
+        valid_schema_name(Prefixes,Id_Ex),
+        error(can_not_insert_class_with_reserve_name(Id), _)),
```
spl, line 3003: This error should be caught, a useful error message should be printed, and a test should be added for that message.

src/cli/main.pl
```diff
@@ -547,6 +547,12 @@
            shortflags([d]),
            default('_'),
            help('document data')],
+          [opt(json),
```
spl, line 550: It would be nice to have CLI tests using this flag.

spl (changes requested):
I've left a number of comments. To add to those, I'm concerned about the introduction of `option/2` and `option/3`. According to the [documentation](https://www.swi-prolog.org/pldoc/man?section=option), each of these “Fails silently if the option does not appear in _OptionList_.” Since some predicates went from having required arguments in their argument list to using `option`, now those predicates will fail when `option` fails. Should we wrap required options with `do_or_die`? Something like `error(unexpected_argument_instantiation(<predicate name>, Options), _))` might help here.

spl:
I like the `option_or_die`. That's a great addition. `merge_options` also removes the problem with failure.

As a minor matter of style, I think adding `insert_documents_default_options` and `delete_documents_default_options` is unnecessary here, because these predicates are used only one time each. I would change this:

```prolog
insert_documents_default_options(
    options{
        graph_type: instance,
        full_replace: false,
        raw_json: false
    }).

api_insert_documents(SystemDB, Auth, Path, Stream, Requested_Data_Version, New_Data_Version, Ids, Options_New) :-
    insert_documents_default_options(Default),
    merge_options(Options_New,Default,Options),
    option(graph_type(Graph_Type), Options),
    option_or_die(author(Author), Options),
    option_or_die(message(Message), Options),
    option(full_replace(Full_Replace), Options),
    option(raw_json(Raw_JSON), Options),
```

into this:

```prolog
api_insert_documents(SystemDB, Auth, Path, Stream, Requested_Data_Version, New_Data_Version, Ids, Options_New) :-
    option(graph_type(Graph_Type), Options, instance),
    option_or_die(author(Author), Options),
    option_or_die(message(Message), Options),
    option(full_replace(Full_Replace), Options, false),
    option(raw_json(Raw_JSON), Options, false),
```

GavinMendelGleason:
I disagree with the stylistic suggestion. I think it's easier to read when not inlined and may be used elsewhere at some point.

src/core/document/json.pl
```diff
@@ -1885,17 +1900,21 @@
...
+    ;   Elt = X^^T,
+        (   C = T % The type is not just subsumed but identical - no ambiguity.
+        ->  value_type_json_type(X,T,V,_)
+        ;   value_type_json_type(X,T,D,T2),
+            % NOTE: We're always compressing, even if Compress_Ids is false
+            % The reason here is that this is a datatype property, not a node of our own.
+            % We may want to revisit this logic though.
+            compress_dict_uri(T2,Prefixes,T2C),
+            V = json{ '@type' : T2C, '@value' : D}
+        )
```
matko, line 1917: could be more defensive to throw here if `C` is not a json and also `Elt` is not a `X^^T`.

src/core/document/json.pl
```diff
@@ -2633,16 +2690,37 @@
...
+    !,
+    database_prefixes(Transaction, Prefixes),
+    do_or_die(
+        (   del_dict('@id', Document, Id, JSON)
+        ->  prefix_expand(Id, Prefixes, Id_Ex),
+            (   is_json_hash(Id_Ex)
+            ->  assign_json_object_id(JSON, Id_Ex)
+            ;   true)
+        ),
+        error(can_not_replace_at_hashed_id(Document), _)),
```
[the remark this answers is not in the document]
spl: This doesn't seem to handled?

spl:
> I disagree with the stylistic suggestion. I think it's easier to read when not inlined and may be used elsewhere at some point.

In that case, I would change all the required options (like `graph_type`) as follows:

```diff
-    option(graph_type(Graph_Type), Options),
+    option_or_die(graph_type(Graph_Type), Options),
```

The latter requires less code tracing to know that `graph_type` is truly required, and it could prevent future errors due to misplaced `graph_type`s.

I'm not asking for this change. I just think it's better to be conservative and defensive about inputs, because that's where problems often arise.

spl (approved):

PR #1227, opened 2022-06-10
Refactor tests to simplify writing them

spl:
* Fix one `SchemaCheckFailure` to put the witness in a list
* Verify success by default
* Simplify verifying failure, not found, etc.
* Use default path to avoid redundant or global variables
* Put the paths, responses, and errors under tests/lib/api
* Remove `agent.defaults()` and use object spread (`...`)


tests/.eslintrc.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
   rules: {
     camelcase: ['error', { properties: 'never', ignoreDestructuring: true }],
     'comma-dangle': ['error', 'always-multiline'],
+    'prefer-object-spread': 'error',
```
matko, line 27: what's this?

spl: If it finds `Object.assign({}, blah)`, it will change it to `{ ...blah }`. See the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).

spl: The most important difference is:

> Note that [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) can be used to mutate an object, whereas spread syntax can't.

matko (approved):

Issue #1254, opened 2022-06-24
Commit hook predicate


matko:
We've been talking about ways to alert other components (like pub/sub infra) that some commit has happened. An easy foundation to do so is to have a dynamic multi-file predicate which is called after every commit, which can then be used to plug in extra behavior into the server.

This should be a fairly easy change and a good way to get started experimenting.

Tails:
Is this a neccessary step for "live query feeds" or "reactive queries", or something else entirely? I guess on the client side with such a hook you can re-run every query to check if anything changed but it would be good to have such a protocol on the DB-level.

matko:
It is a necessary step both for 'live query feeds' and 'reactive queries, but it implements neither. It is just a hook, an extra opportunity to inject some prolog code to run at a particular point in the process. The hook, as now implemented in the draft pull request, runs a prolog predicate on the server right after commit, in a context where that commit is still fully in memory and can be analyzed by the hook. So that should be a good basis to implement any sort of reactiveness.

The most basic kind would just be to wake up a client for everything and let them do a query to find out if anything interesting happened, but as you're pointing out, it's much better if somehow the DB did this. You don't want to potentially wake up a very large number of clients for changes that don't interest them. Instead, it would make sense to me if there was some mechanism by which clients registered what they were interested in (maybe submitting a query? Or maybe a less granular topic? I don't know), and the server only woke up interested clients.

A further complicating factor is that you might be running not just one terminusdb instance, but a large cluster of them, and changes could be coming from any one of them. A simple hook doesn't account for the cluster situation at all. Some bigger orchestration will be needed for that.

So yeah, I think this hook is the foundation for further experimenting in building reactive systems, but it's really just that. There's a lot of figuring out to do still before we can truly have 'live query feeds', 'reactive queries', or whatever else. But at least it'll be possible to experiment with solutions like that to figure out what actually works.

hoijnet:
Just making a webhook request with a GRAPH_SPEC or similar would be such an awesome thing, this is an awesome foundation to build upon!

alexander-mart:
@matko @hoijnet @Tails 

It already implemented for self-hosted docker version with this plugin:
https://github.com/terminusdb-labs/terminusdb-plugins/blob/main/webhook.pl

How to add this plugin:
https://github.com/terminusdb-labs/terminusdb-plugins/blob/main/README.md

PR #1255, opened 2022-06-24
A rudimentary plugin architecture

matko:
This pull request introduces a rudimentary plugin architecture, which will load plugin files from the storage/plugins directory.

This implements #1254.


src/core/transaction/database.pl
```diff
@@ -275,8 +275,8 @@
     commit_validation_objects(Validations, Committed),
     collect_validations_metadata(Validations, Validation_Meta_Data),
     collect_commit_metadata(Committed, Commit_Meta_Data),
-    put_dict(Validation_Meta_Data, Commit_Meta_Data, Meta_Data).
-
+    put_dict(Validation_Meta_Data, Commit_Meta_Data, Meta_Data),
+    ignore(foreach(post_commit_hook(Validations, Meta_Data), true)).
```
GavinMendelGleason, line 279: Not sure if this needs to be `foreach` rather than `forall`. What things are going to become more bound given that Validations and Meta_Data are already bound?

GavinMendelGleason (approved):

src/core/transaction/database.pl
matko: You're right, it should just all be forall.

GavinMendelGleason: or it could be `exhaust` from utils.pl

Issue #771, opened 2021-11-15
Users and Organizations API updates


Francesca-Bit:
Operations desired:
* Create a user
* Delete a user
* Create an organization
* Delete an organization
* Add a user to an organization
* Remove a user from an organization
* Get all organizations for a user
  * Get all organizations for the authenticated user
* Get all databases to which a user has access
  * Get all databases to which the authenticated user has access

spl:
See also <https://discuss.terminusdb.com/t/ganymede-sustainable-food/280/5>:

<blockquote>
The biggest issue I’m dealing with right now is the changes in version 10. I’m mostly focused on simplifying (automating) the deployment (with Docker) for each organisation. Other issues are lower down on my list.

I’m having to infer what best practice is for setting up an organisation on an instance based on issues, comments, and code, as the docs seem to be missing that key part, so any advice or guidance on what we should be looking for there would be good.

The docs cover changing organisation names, adding users to organisations, etc. Adding an organisation and authenticating / role management is less clear.

My understanding is that this has to be done with the default user (admin) in the first instance, currently using authentication through the bearer token; I have a OIDC / OAuth2 solution for long-term, but that’s outside of the scope of the experiments we’re currently running against it.
</blockquote>

spl:
I've updated the summary. Here are @Francesca-Bit's previous comments:
- get all the organizations and roles (organization level) of a current user (`/api/organizations`)
- get all the databases and roles (databases level) of a current user, I was thinking that we'll need roles related with databases too (`/api/organizations/:orgid/databases`)

Francesca-Bit:
This function will return the list of organizations and databases
terminusdb-client-js:
tmpClient.getUserOrganizations().then(result=>{console.log})
the result will be like this 
[
  {
    "@id":"Organization/my_team",
    "@type":"Organization",
    "databases": [
      {
	"@id":"UserDatabase/5efbe4c00aabad2738b332cb3db306c8d49779eda7319aa3af3ee9....",
	"@type":"UserDatabase",
	"comment":"test",
	"creation_date":"2022-01-13T15:05:53.337Z",
	"label":"test",
	"name":"test",
	"state":"finalized"
      }
    ],
    "name":"my_team"
  }]

spl:
This could include <https://github.com/terminusdb/terminusdb/issues/954>.

GavinMendelGleason:
Fixed in #1265

PR #1265, opened 2022-06-29
User endpoints

GavinMendelGleason:


GavinMendelGleason:
Addresses #771

src/core/api/api_access_control.pl
```diff
@@ -0,0 +1,486 @@
...
+        (   t(Id,rdf:type,'@schema':'Organization'),
+            get_document(Id,Organization))).
+
+api_add_organization(_, Auth, Organization, Id) :-
+    do_or_die(
+        is_super_user(Auth),
+        error(access_not_authorised(Auth,'Action/manage_capabilities','SystemDatabase'), _)),
+
+    % Make sure that we only insert Organization types.
+    put_dict('@type', Organization, 'Organization', Organization1),
```
matko, line 190: I don't think it's a good idea to add Organization into the system database after making slight changes to the document. While you blacklist the database field here, you're not blacklisting whatever other fields we may come up with in the future, and it'll be hard in future reviews to remember that there's this place in the code that needs to be changed to make sure people don't just insert pre-filled records into our database.

Better to start with a fresh document and fill it in, rather than take things out.

src/core/api/api_access_control.pl
```diff
@@ -0,0 +1,486 @@
...
+    ;   throw(error(no_unique_id_for_role_name(RoleName),_))
+    ).
+
+api_add_role(_, Auth, Role, Id) :-
+    do_or_die(
+        is_super_user(Auth),
+        error(access_not_authorised(Auth,'Action/manage_capabilities','SystemDatabase'), _)),
+
+    % Make sure that we only insert role types.
+    put_dict('@type', Role, 'Role', Typed_Role),
```
matko, line 94: I don't think we should just insert the Role document (with added type), due to concerns around data validation. As a general principle I think we should be very very suspicious of any incoming user data, especially if these are records that have to go into the system database. Instead of just inserting them directly (after slight modifications), I'm in favor of always building these sorts of documents ourselves, filling out details using the input document. This ensures we take a whitelist approach to what goes into our database, rather than just trusting the user to not submit documents with extra unexpected fields, or blacklisting particular fields we don't trust.

src/core/api/api_access_control.pl
```diff
@@ -0,0 +1,486 @@
...
+        _
+    ).
+
+api_update_role(_, Auth, Role) :-
+    do_or_die(
+        is_super_user(Auth),
+        error(access_not_authorised(Auth,'Action/manage_capabilities','SystemDatabase'), _)),
+
+    % Make sure that we only insert role types.
+    put_dict('@type', Role, 'Role', Typed_Role),
```
matko, line 110: Same concern here regarding treating input data as documents to put into the system database. See above.

src/core/api/api_access_control.pl
```diff
@@ -0,0 +1,486 @@
...
+    ->  with_transaction(
+            System_Context,
+            (   get_document(System_Context, Capability_Id, Capability),
+                (   get_dict('role', Capability, Old_Role_Ids)
+                ->  true
+                ;   Old_Role_Ids = []),
+                append(Role_Ids,Old_Role_Ids, New_Role_Ids),
+                sort(New_Role_Ids,Sorted_Roles),
+                put_dict('role',Capability,Sorted_Roles, New_Capability),
+                replace_document(System_Context,New_Capability)
```
matko, line 271: a `replace_field` predicate would be really useful here...

matko (changes requested):
Very happy to see this change finally coming through.
My comments mostly deal with data validation in api_access_control.pl. I think we should be more rigorous in never treating input documents as documents that go into the system database, as this is inherently risky. See the comments in that file for details.

src/core/api/api_access_control.pl
```diff
@@ -0,0 +1,486 @@
...
+
+    create_context(SystemDB, commit_info{author: "admin", message: "API: Add Organization"},
+                   System_Context),
+    with_transaction(
+        System_Context,
+        insert_document(System_Context,New_User,Id),
+        _
+    ).
+
+api_delete_user(_, Auth, User_Id) :-
```
matko, line 373: This should make sure you're not deleting admin.

src/core/api/api_error.pl
```diff
@@ -1068,6 +1068,130 @@
                               'api:key' : Key},
              'api:message' : Msg
             }.
+api_error_jsonld_(role,error(no_unique_id_for_role_name(Name),_), JSON) :-
+    format(string(Msg), "There is either more than one id for role ~s. Consider deleting duplicates if you want to refer to them by name rather than id.", [Name]),
```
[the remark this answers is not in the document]
GavinMendelGleason: Yeah, copy paste bug. It used to catch two different errors.

src/cli/main.pl
```diff
@@ -1426,6 +1615,271 @@
...
+    (   option(json(true),Opts)
+    ->  json_write_dict(current_output,Organizations,[])
+    ;   forall(member(Organization,Organizations),
+               (   get_dict(name,Organization,Organization_Name),
+                   get_dict('@id',Organization,Organization_Id),
+                   format(current_output, "'~s' has id: '~s'~n",
+                          [Organization_Name,Organization_Id])
+               ))
+    ).
+run_command(user,create,[Name], Opts) :-
```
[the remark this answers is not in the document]
GavinMendelGleason: Actually I think we need this now.

src/cli/main.pl
```diff
@@ -778,6 +779,194 @@
...
+           default(false),
+           help('Interpret argument as an organization id rather than a name.')],
+          [opt(json),
+           type(boolean),
+           longflags([json]),
+           shortflags([j]),
+           default(false),
+           help('Return answer as a JSON document')]]).
+opt_spec(user,create,'terminusdb user create USER',
+         'Create a user with a given name USER in the existing organization ORGANIZATION',
```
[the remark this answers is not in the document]
GavinMendelGleason: Indeed! Because it doesn't use one. :D

matko (approved):

PR #1270, opened 2022-07-03
Send query output to stdout

GavinMendelGleason:


rrooij (approved):

rrooij:
Tested locally and it now correctly prints to stdout.

Issue #1237, opened 2022-06-16
Make JWT subject claim name configurable, sensible or with good fallback


hoijnet:
**Is your feature request related to a problem? Please describe.**
Configuring JWT authentication in TerminusDB is harder than it should be. The default claim name (`http://terminusdb.com/schema/system#agent_name`) is non-standard and can be very hard to configure at an IdP without having specific skill and an awful lot of permissions; which could be hard to gain in an enterprise context.

Making the default claim name to use from the JWT token would make this much easier. The straight proposal would be to have a fallback to `preferred_username` and if that does not exist, to the `sub` default unique subject name if the others don't exist.

The fix would likely be straightforward, and solve the most important aspect of integration in this older issue:
* [Inconsistent behaviour for JWT JWKS keys/certificates endpoints](https://github.com/terminusdb/terminusdb/issues/728)

The specific code location referenced:
[routed.pl#L2893](https://github.com/terminusdb/terminusdb/blob/dd52f85caa39fc7a89310b5e63ec1902720bf0ad/src/server/routes.pl#L2893)

**Describe the solution you'd like**
I would like to accept a token that includes `preferred_username` with the user to use in TerminusDB, and failing that, fall back to `sub`. Even better would be to make it configurable as that would enable me to use any of the claims in the token, and make the behaviour very clear. 

**Describe alternatives you've considered**
One alternative would be to add the claim to the IdP, which is often not even possible to do without being superadmin. The workaround to use header-based authentication is not strong enough in my opinion, using a JWT is a better and stronger alternative.

**Additional context**

rrooij:
Solved in: 68a9788d15e7b3394be17121d3faf61f3aa86834 I hope this works for you! :smile:

hoijnet:
Great, this will simplify things a lot! For those seeing this, could be interesting to look at issue #728 as well for context. Thanks for providing the option, this will be super-helpful for integrating TerminusDB!

PR #1274, opened 2022-07-07
Add configurable claim name for JWT tokens

rrooij:
Introduce a new ENV variable called TERMINUSDB_JWT_SUBJ_CLAIM_NAME. By default it is set to the `http://terminusdb.com/schema/system#agent_name` but it can be configured by the user to mean something else as well.

This should solve: https://github.com/terminusdb/terminusdb/issues/1237

### DIFF
--- a/src/config/terminus_config.pl
+++ b/src/config/terminus_config.pl
@@ -111,7 +111,7 @@ jwt_jwks_endpoint(Endpoint) :-
     ;   Endpoint = Value).
 
 jwt_subject_claim_name(Name) :-
-    getenv_default('TERMINUSDB_JWT_SUBJ_CLAIM_NAME', 'http://terminusdb.com/schema/system#agent_name', Name).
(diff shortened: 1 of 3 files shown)


src/config/terminus_config.pl
```diff
@@ -109,6 +110,9 @@
     ->  false
     ;   Endpoint = Value).
 
+jwt_subject_claim_name(Name) :-
+    getenv_default('TERMINUSDB_JWT_SUBJ_CLAIM_NAME', 'http://terminusdb.com/schema/system#agent_name', Name).
```
matko, line 114: please write out `SUBJECT`. There is no need to save on 3 letters here.

matko, line 114: Is this a good standard? It is what we do for TerminusX but since it's non-standard maybe it's better to default to something that is normally in a JWT already?

rrooij:
Only the python tests fail. That's because it has not yet been configured to a different header. If we decide to merge this one, I can merge this one first as well: https://github.com/terminusdb/terminusdb-client-python/pull/315

matko (approved):

PR #1280, opened 2022-07-11
Fix snap and non-docker tests by creating subdirs as well

rrooij:
The build fails otherwise. This change is needed because make_directory does not make subdirs. The storage directory is not always there already when running the tests.


matko:
If so, shouldn't we change these tests to create a slightly less weird test dir? There is no reason for the testdir to be in storage/ right?

rrooij:
As you can see, /tmp/ fails.

rrooij:
Okayish? Or do we want to remove `make_directory_path` since we know that `tmp` will always exist?

PR #1320, opened 2022-07-22
Allow branches to be diffed directly, and small bugfix in uniqueness

matko:
There was a small bug where a full removal of a document didn't properly bind the ID for that deletion.

Also, it should now be possible to directly specify a branch to diff, instead of a commit id.

### DIFF
--- a/docs/terminusdb.1.ronn
+++ b/docs/terminusdb.1.ronn
@@ -267,10 +267,10 @@ or two commits (path required).
   document id to use for comparisons
 
   * `-p`, `--before_commit`, `--before-commit`=[value]:
-  Commit of the *before* document(s)
+  Commit or branch of the *before* document(s)
 
   * `-s`, `--after_commit`, `--after-commit`=[value]:
-  Commit of the *after* document(s)
+  Commit or branch of the *after* document(s)
 
 ### apply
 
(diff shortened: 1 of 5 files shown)


rrooij:
Looks good! But can you add an integration/unit test for this as well?

rrooij:
Does not work for me when creating a branches and committing a document to the `foo` one:

`./terminusdb diff admin/test/local/branch/foo admin/test/local/branch/main`

Gives the help screen:

```

terminusdb diff [Path] OPTIONS

Create a diff between two JSONs, a JSON and a commit (path required),
or two commits (path required).

--help           -h  boolean=false                      print help for the `diff` command
--before         -b  atom=_                             JSON document which is the *before*
--after          -a  atom=_                             JSON document which is the *after*
--keep           -k  atom={"@id" : true, "_id" : true}  Skeleton of the document to retain as
                                                          context
--copy-value,
--copy_value     -c  boolean=false                      Maintain explit copies of diffs in lists
--docid          -d  atom=_                             document id to use for comparisons
--before_commit,
--before-commit  -p  atom=_                             Commit of the *before* document(s)
--after_commit,
--after-commit   -s  atom=_                             Commit of the *after* document(s)
```

matko:
The invocation should be
```
./terminusdb diff admin/test --before_commit=main --after_commit=foo
```

rrooij (approved):

PR #1361, opened 2022-08-11
Adding inheritence check

### LINKED ISSUE #776: Schema checking does not verify that all inherited classes exist
Try creating a schema which inherits from a non-existing class. This appears to be accepted but is not a valid schema.

GavinMendelGleason:
Fixes  #776


rrooij (approved):

rrooij (changes requested):
Do the enums as well

rrooij (approved):

PR #1386, opened 2022-08-19
Meta tag

### LINKED ISSUE #668: Schema: Storing arbitrary class/property metadata
**Is your feature request related to a problem? Please describe.**
I have data that describe classes and properties which I want to store in my schema.

Use cases:
- Importing data from external sources (e.g. storing that a class `Person` is equivalent to `wd:Q5` and its property `scm:birthplace` to `wdt:P19` on Wikidata)
- Storing colours or icons related to classes and properties which can be used in an frontend application

**Describe the solution you'd like**
Something similar to `@documentation` – maybe called `@metadata`:
```js
{
  "@id": "Person",
  "@type": "Class",
  "@documentation": {
    "@comment": "A person",
    "@properties": {
      "place_of_birth": "The persons birthplace"
    }
  },
  "@metadata": {
    "wikidata_class": "wd:Q5",
    "color": "indigo"
  },
  "place_of_birth": {
    "@type"  : "Optional",
    "@class" : "Place",
    "@metadata": {
      "wikidata_property": "wdt:P19",
      "icon": "map.png"
    }
  }
}
```

**Describe alternatives you've considered**
The above proposal would be quite easy to use and – i imagine – not that hard to implement either. But I also see some value for more structured approaches (i.e. letting users define a set of properties classes and properties can have).

One option would be to create `ClassMetadata` and `PropertyMetadata` classes in the schema and store the arbitrary metadata in the instance graph. This is almost already possible, the only issue being that it doesn't seem to be possible to have a _link property_ that links to the type `owl:Class` (if I don't want to specify the class a `ClassMetadata` document is referring to as a string).

In my eyes it doesn't really make sense to store that sort of metadata on the instance graph. Instead I could imagine – no idea if this is technically feasible so this is purely speculative – a schema of a schema which would allow me to define custom `@`properties, maybe prefixed with `@meta_`.
```js
{
  "@id": "Person",
  "@type": "Class",
  "@documentation": {
    "@comment": "A person",
    "@properties": {
      "place_of_birth": "The persons birthplace"
    }
  },
  "@meta_wikidata_class": "wd:Q5",
  "@meta_color": "indigo"
  "place_of_birth": {
    "@type"  : "Optional",
    "@class" : "Place",
    "@meta_wikidata_property": "wdt:P19",
    "@meta_icon": "map.png"
  }
}
```

### LINKED ISSUE #669: Schema object documentation annotations
**Is your feature request related to a problem? Please describe.**

I am missing what, in the previous version, existed as a label on schema elements; to enable a flexible description of the schema object.

To bring the fluidity of versioned code to versioned data; most things should be mutable–including schema element names that may change as the data product evolves. I had envisioned using random ids for schema elements (such as document and subdocument schema definitions), and use labels for the names of them. 

Most schema changes happen early in defining a model, but sometimes it is hard to get the wording right immediately before understanding the domain fully; and it would make sense with a fluid label.

To have a label for the schema solves this as it can be changed later, whereas having an id would lead to continuous use of the old (incorrect) term for something in evolving schemas. In code, refactoring is an option - I could make the argument it would help here too (if one wants to). The description document element is used for a textual description of the schema, and could be coerced into a json to be a more rich description of the schema/which is could ingestable for computers too, but it would be to misuse the field.

What could be helpful for schema implementers (as in my case), is schema annotation, where the implementer would be choosing what kind of schema annotations would be necessary for their specific use cases. In my case, label is helpful, someone else might have specific properties they want to annotate a class with (such as a connected query type or name of it, others might want to store class i18n label names for internationalization of their classnames, or just a unit of measure for values contained in a subdocument, unsure if I would model it that way though). 

The example of the WOQL schema below illustrates the point; it seems you needed specific annotations on the class for the AST and the name of it in the case of the named Query, you might have other cases with different needs.

```
{ "@id" : "NamedQuery",
  "@type" : "Class",
  "@documentation" : {
      "@comment" : "A named query names a specific query for later retrieval and re-use.",
      "@properties" : { 
                        "name" : "The name of the NamedQuery to be retrieved",
                        "query" : "The query AST as WOQL JSON" }
              },
  "@key" : { 
             "@type" : "Lexical",
             "@fields" : [ "name" ] 
         },
  "name" : "xsd:string",
  "query" : "Query" }
```

A more free-form key such as "@properties" above with a flat namespace as shown would be a more flexible approach for schema annotation, whilst not breaking the strict schema checking, and enabling to keep the metadata within the data product itself at all times.

I believe the choice should be there for adding an optional label, and that an @documentation item called label would be helpful to have; for those that want to use labels for their schema elements, where id and label differ, or making the transition.

Additional considerations for this feature includes i18n (internationalisation) of the label (and perhaps of the description and each property as well), and the ability to add other related annotations, that could be both machine-orientated, and human-orientated, for provenance, integration instructions, property validation/sanitisation/allowable values, and other needs of data model developers. 

The "@documentation"/"@properties" as a JSON object notation (with fairly free naming of the properties inside) with @comment being a valid optional part used in the Product Model UI as you do now, looks neat as done in the WOQL Schema. 

An alternative would be to add an "@annotation"/{} JSON object side by side of "@documentation", or as a property in "@documentation". See the solution descriptions below.

**Describe the solution you'd like**

In order of preference, the following solutions could make sense:

1. Add an @annotation namespace (next to the @documentation namespace) with free-form key/value pairs for the document/subdocument definition that can hold arbitrary information about the document/subdocument/enum, its properties or other relationship with the rest of the world
2. Add an @annotation key/value object within @documentation (same as above)
3. Extend the @documentation element to support arbitrary key/value pairs, including @label, @description, and any other key/value pair (unsure if this is a good idea though, just a top-of-mind idea)
4. Add @label/title as an allowable entry added to the @documentation entry

**Describe alternatives you've considered**
A workaround is to put a JSON object into the description field that contains both label and description key/values.

**Additional context**

GavinMendelGleason:
Fixes #669 and Fixes #668

### DIFF
--- a/src/core/document/schema.pl
+++ b/src/core/document/schema.pl
@@ -23,6 +23,8 @@
               schema_key_descriptor/4,
               documentation_descriptor/3,
               schema_documentation_descriptor/3,
+              metadata_descriptor/3,
+              schema_metadata_descriptor/3,
               oneof_descriptor/3,
               schema_oneof_descriptor/3,
               type_family_constructor/1,
@@ -57,6 +59,7 @@
 
 :- use_module(json). % This feels very circular.
 :- use_module(instance). % This is most definitely circular.
+:- use_module(json_rdf, [graph_get_json_object/3]).
 
 graph_member_list(Instance, O,L) :-
     xrdf(Instance, L, rdf:first, O).
@@ -78,11 +81,11 @@ is_choice(Validation_Object, Class) :-
     database_schema(Validation_Object,Schema),
     is_schema_choice(Schema, Class).
 
-:- table is_schema_enum/2 as private.
+%:- table is_schema_enum/2 as private.
 is_schema_enum(Schema, Class) :-
     xrdf(Schema, Class, rdf:type, sys:'Enum').
 
-:- table is_schema_choice/2 as private.
+%:- table is_schema_choice/2 as private.
 is_schema_choice(Schema, Class) :-
     xrdf(Schema, Class, rdf:type, sys:'Choice').
 
@@ -90,7 +93,7 @@ is_foreign(Validation_Object,Class) :-
     database_schema(Validation_Object,Schema),
     is_schema_foreign(Schema, Class).
 
-:- table is_schema_foreign/2 as private.
+%:- table is_schema_foreign/2 as private.
 is_schema_foreign(Schema, Class) :-
     xrdf(Schema, Class, rdf:type, sys:'Foreign').
 
@@ -98,7 +101,7 @@ is_tagged_union(Validation_Object,Class) :-
     database_schema(Validation_Object,Schema),
     is_schema_tagged_union(Schema, Class).
 
-:- table is_schema_tagged_union/2 as private.
+%:- table is_schema_tagged_union/2 as private.
 is_schema_tagged_union(Schema, Class) :-
     xrdf(Schema, Class, rdf:type, sys:'TaggedUnion').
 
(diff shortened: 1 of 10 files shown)


src/core/document/schema.pl
```diff
@@ -376,6 +388,7 @@
 refute_schema(Validation_Object,Witness) :-
     database_prefixes(Validation_Object, Prefixes),
     is_simple_class(Validation_Object,Class),
+    \+ is_json_class(Validation_Object, Class),
```
matko, line 391: So if I understand correctly, as long as something is tagged as a json class, we just believe that?
Isn't there some more elaborate check to be done to make sure this json object doesn't contain weirdness that should not be part of it?

GavinMendelGleason: It's possible that you could confuse the schema by direct triple load, but I don't think it's possible by loading documents.

matko (approved):

hoijnet:
I am so looking forward to exposing and making use of this functionality soon 🎉

Issue #1395, opened 2022-08-22
openapi: log not documented


rrooij:
The `/api/log` endpoint is not documented but very useful. We should add it.

It is not yet available in `main`: https://editor.swagger.io/?url=https://raw.githubusercontent.com/terminusdb/terminusdb/main/docs/openapi.yaml

matko:
it is now documented.

rrooij:
Solved in: https://github.com/terminusdb/terminusdb/pull/1396 Just for future reference

PR #1396, opened 2022-08-22
Add log endpoint openapi

rrooij:
See issue #1395


GavinMendelGleason (commented):
Needs a documentation string so that it will generate readable documentation from description

GavinMendelGleason (approved):

PR #1410, opened 2022-08-25
Allow loading of trig files

GavinMendelGleason:

### DIFF
--- a/src/core/triple/literals.pl
+++ b/src/core/triple/literals.pl
@@ -349,7 +349,7 @@ reset_normalise_warning :-
 normalise_triple(rdf(X,P,Y,G),rdf(XF,P,YF)) :-
     (   normalise_warning_
     ->  true
-    ;   json_log_error_formatted('Warning: ignoring graph ~q~n', [G]),
+    ;   json_log_warning_formatted('Warning: ignoring graph ~q~n', [G]),
         set_normalise_warning
     ),
     normalise_triple(rdf(X,P,Y),rdf(XF,P,YF)).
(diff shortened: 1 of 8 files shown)


rrooij:
Can you provide a simple integration test? :smile: Something very simple like inserting a `trig` in a schemaless DB.

src/core/triple/literals.pl
```diff
@@ -337,6 +338,21 @@
...
+set_normalise_warning :-
+    assertz(normalise_warning_).
+
+reset_normalise_warning :-
+    retractall(normalise_warning_).
+
+normalise_triple(rdf(X,P,Y,G),rdf(XF,P,YF)) :-
+    (   normalise_warning_
+    ->  true
+    ;   format(user_error,'Warning: ignoring graph ~q~n', [G]),
```
matko, line 352: Please use `json_log_warning_formatted` instead so it works properly with different logging backends we set up.

matko (changes requested):

rrooij:
Made a simple integration test

src/core/triple/literals.pl
```diff
@@ -337,6 +338,21 @@
...
+set_normalise_warning :-
+    assertz(normalise_warning_).
+
+reset_normalise_warning :-
+    retractall(normalise_warning_).
+
+normalise_triple(rdf(X,P,Y,G),rdf(XF,P,YF)) :-
+    (   normalise_warning_
+    ->  true
+    ;   json_log_error_formatted('Warning: ignoring graph ~q~n', [G]),
```
matko, line 352: please use `json_log_warning_formatted` as it is not an error.

matko (changes requested):

matko (approved):

PR #1424, opened 2022-09-01
Oneline JSON for all requests

rrooij:
All requests should return a oneline JSON for performance reasons and easier parsing. Pretty printing can optionally be done by other tools, such as `jq`. It also reduces a small bit of network bandwidth.


rrooij:
Should we do this for the CLI as well?

GavinMendelGleason (approved):

PR #1439, opened 2022-09-05
Add test for database uri

### LINKED ISSUE #1422: Create database api does not return the db id
after creating a new database, we get the following success message
`{@type: "api:DbCreateResponse"
api:status: "api:success"}`,
it would be more useful if the api returns the database id

GavinMendelGleason:
Fixes #1422

### DIFF
--- a/tests/test/db-auth.js
+++ b/tests/test/db-auth.js
@@ -20,7 +20,8 @@ describe('db-auth', function () {
 
   it('passes create with no comment', async function () {
     const bodyString = '{"label":"hello"}'
-    await db.create(agent, { bodyString })
+    const result = await db.create(agent, { bodyString })
+    expect(result.body).to.have.property('api:database_uri')
     await db.delete(agent)
   })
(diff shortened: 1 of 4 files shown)


tests/test/db-auth.js
```diff
@@ -20,7 +20,8 @@ describe('db-auth', function () {
 
   it('passes create with no comment', async function () {
     const bodyString = '{"label":"hello"}'
-    await db.create(agent, { bodyString })
+    const result = await db.create(agent, { bodyString })
+    expect(result.body).to.have.property('api:database_uri')
```
rrooij, line 24: Can't you check whether the URL is correct as well?

rrooij (changes requested):

tests/test/db-auth.js
GavinMendelGleason: It is!

rrooij (approved):

PR #1447, opened 2022-09-07
Json diff for top level `sys:JSONDocument`

### LINKED ISSUE #1394: Apply does not work on JSONDocuments at top level
The following does not work because the document complains that it can not infer the document type. This is because it *must* use a raw_json flag on insertion, which must likewise be reported by the diff. 

```shell
# Create Database
terminusdb db create admin/foo
# Add a JSON document
echo '{ "name": "Socrates" }' | terminusdb doc insert admin/foo -j
# Get the document back
terminusdb doc get admin/foo # Prove that it works, Socrates is there!
# Branch the database
terminusdb branch create admin/foo/local/branch/changes --origin admin/foo
# Add some more philosophers
echo '{ "name": "Plato" }' | terminusdb doc insert admin/foo/local/branch/changes -j
echo '{ "name": "Aristotle" }' | terminusdb doc insert admin/foo/local/branch/changes -j

# Look at the difference in the databases
terminusdb diff admin/foo --before-commit main --after-commit changes | jq

# Apply the differences

terminusdb apply admin/foo --before-commit main --after-commit changes | jq
```

GavinMendelGleason:
Fixes #1394

### DIFF
--- a/src/core/api/api_patch.pl
+++ b/src/core/api/api_patch.pl
@@ -60,7 +60,12 @@ document_from_commit(Branch_Descriptor, Commit_Id, Doc_Id, Document, Transaction
         open_descriptor(Commit_Descriptor, commit_info{}, Transaction, Map_In, Map_Out),
         error(unresolvable_collection(Commit_Descriptor),_)),
 
-    get_document(Transaction, Doc_Id, Document).
+    Options = options{
+                  compress_ids : true,
+                  unfold: true,
+                  keep_json_type: true
+              },
+    get_document(Transaction, Doc_Id, Document, Options).
 
 api_diff_id(System_DB, Auth, Path, Before_Version, After_Version, Doc_Id, Diff, Options) :-
     resolve_descriptor_auth(read, System_DB, Auth, Path, instance, Branch_Descriptor),
--- a/src/core/document/json.pl
+++ b/src/core/document/json.pl
@@ -8,6 +8,7 @@
               json_schema_triple/3,
               json_schema_elaborate/3,
               get_document/3,
+              get_document/4,
               get_document/5,
               get_document_uri/3,
               get_schema_document/3,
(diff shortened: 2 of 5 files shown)


src/core/api/api_patch.pl
```diff
@@ -64,7 +64,12 @@
         open_descriptor(Commit_Descriptor, commit_info{}, Transaction, Map_In, Map_Out),
         error(unresolvable_collection(Commit_Descriptor),_)),
 
-    get_document(Transaction, Doc_Id, Document).
+    Options = options{
+                  compress_ids : true,
+                  unfold: true,
+                  keep_json_type: true
```
matko, line 70: Should this also be implemented in the rust version?

src/core/document/json.pl
```diff
@@ -570,6 +572,10 @@
             select_dict(json{'@id' : _ }, Context_With_ID, Context)
         ;   Context = _{})
     ).
+/*
+    default_prefixes(Prefixes),
+    put_dict(Context,Prefixes,Expanded).
+*/
```
matko, line 578: might as well remove these comments

src/core/document/json.pl
```diff
@@ -586,6 +592,11 @@
     create_context(Askable, Query_Context),
     database_prefixes(Query_Context, Context).
 
+database_and_default_prefixes(Askable, Prefixes) :-
+    database_prefixes(Askable, Database_Prefixes),
+    default_prefixes(Default_Prefixes),
```
matko, line 597: if we're gonna have a set of default prefixes, we should probably enforce that those prefixes cannot be used by users in their own graphs. Do we do that?

GavinMendelGleason: I don't think that is necessary. If someone uses a prefix to mean something else, and they submit something else, that's their problem not ours.

matko (approved):

PR #1454, opened 2022-09-09
Error when retrieving log on something with no history

matko:
This ensures that we only attempt to build a log from things that actually have logs (branch and commit descriptors).


matko:
Looks like retrieving a log from a particular commit still errors, so this remains a draft.

GavinMendelGleason (approved):

PR #1457, opened 2022-09-09
Fix captures in full replace

### LINKED ISSUE #1398: Strange error on insertion with captures
Using this schema: https://github.com/terminusdb-labs/alice_concordance/blob/main/schema/concordance.json

We get an error with the following insert: 

```shell
echo '{
    "@type": "Paragraph",
    "@capture": ".paragraph 1 0",
    "text": "Down the Rabbit-Hole",
    "terms": [
      {
        "@type": "TermDF",
        "term": {
          "@ref": ".term Down"
        },
        "df": 1
      }
    ]
  },
  {
    "@type": "Term",
    "@capture": ".term Down",
    "tf": 1,
    "term": "Down"
  }' |  ./terminusdb doc insert admin/alice

```

matko:
When doing a full replace on instance data, the insert logic was not properly suspending inserts until all captures were known, causing the code to attempt to insert unbound triple data. This led to an ugly stack trace.

Fixes #1398.

### DIFF
--- a/src/core/api/api_document.pl
+++ b/src/core/api/api_document.pl
@@ -201,37 +201,43 @@ api_insert_document_(instance, Raw_JSON, Transaction, Document, state(Captures_I
         do_or_die(insert_document(Transaction, Document, Raw_JSON, Captures_In, Id, _Dependencies, Captures_Out),
                   error(document_insertion_failed_unexpectedly(Document), _))).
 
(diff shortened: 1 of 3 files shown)


matko:
On second though, this doesn't fix #1398, only the secondary issue that was also reported in that issue thread, but which is kinda of its own issue.

GavinMendelGleason (changes requested):
I co-developed this so I believe it is correct, but we need a test to avoid regressions.

GavinMendelGleason (approved):

PR #1473, opened 2022-09-16
change lazy doc logic to have the attr unify hook unify on head

matko:
I think it is better if hooks are able to unify on head, because this allows swipl to use indexing and hooks will be called quite often for anything that has attributes, not just our lazy doc lists.

So this will prevent our unify hook for lazy doc lists from being called on everything else.


matko:
On second thought, and rereading the docs a bit more carefully, the fact that the attribute is added under a module already limits the scope of attribute hook execution, and this change is unnecessary.

PR #1477, opened 2022-09-19
reverse linking of subdocs

matko:


rrooij (approved):
Tested extensively. Works very well but has performance issues when inserting subdocuments first and then the parent document.

PR #1484, opened 2022-09-23
Add signal handling for SIGINT

GavinMendelGleason:

### DIFF
--- a/src/cli/main.pl
+++ b/src/cli/main.pl
@@ -21,8 +21,10 @@
 :- use_module(library(optparse)).
 :- use_module(core(util),
               [do_or_die/2, token_authorization/2,
+               json_log_notice/1,
                basic_authorization/3, intersperse/3,
-               with_memory_file/1, with_memory_file_stream/3]).
+               with_memory_file/1, with_memory_file_stream/3,
+               json_log_notice_formatted/2]).
 :- use_module(core(plugins)).
 :- use_module(library(prolog_stack), [print_prolog_backtrace/2]).
 :- use_module(library(apply)).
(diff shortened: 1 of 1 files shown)


rrooij (approved):
Looks good! We might need to think about creating a cleanup job if there are things that should be saved before halting.

rrooij (changes requested):
When testing this and running `terminusdb serve` and pressing Ctrl+C, I get the following error:

```
./terminusdb serve                                                                                                                                         signal_handler 1d ● ⬡
% Started server at http://localhost:6363/
% TerminusDB server started at Sunday, Sep 25, 20:44:32 CEST (utime 1664131472.9562778) args [serve,help(false),interactive(false),memory(_8906)]
% Welcome to TerminusDB's terminusdb-server, version 10.1.5!
% You can view your server in a browser at 'http://localhost:6363'

^CError: error(resource_error(stack),stack_overflow{choicepoints:10,depth:31,environments:23,globalused:50,localused:11,stack:[frame(31,system:call((:)/2),[]),frame(30,system:reset((:)/2,_190,_192),[]),frame(29,'$tabling':delim(ret/1,(:)/2,94475679990480,[]),[]),frame(28,'$tabling':activate(ret/1,(:)/2,94475679990480),[]),frame(27,'$tabling':run_leader(ret/1,(:)/2,fresh/2,_292,_294),[])],stack_limit:16201108,trailused:1})
```

This is regardless of the storage being used.

GavinMendelGleason:
This change will not alter the behaviour in `-i` since this starts a toplevel which has its own signal handlers, but instead fixes the bug in trying to interupt in a shell. 

It does not, however, give me the same behaviour of a stack-overflow... curious how that happens.

rrooij:
Perhaps it is my SWI Prolog version. I'm using v8.4.2.

rrooij:
Same problem with SWI Prolog v8.4.3 unfortunately.

PR #1490, opened 2022-09-29
Upgrade SWIPL version

rrooij:
Upgrade SWI Prolog version to v8.4.3 (current stable) from v8.4.2. Lets make sure that all the tests pass and that the password_hash is still backwards compatible.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ARG DIST=community
 
 # Install the SWI-Prolog pack dependencies.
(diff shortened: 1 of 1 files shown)


matko (approved):
This will probably be unproblematic for us, but we know there's a bug in the crypto initialization for the 8.4 branch which just didn't happen to manifest in the compiles so far. I was hoping we could wait for that fix to be backported, but it looks like it's not going to happen any time soon.
So let's just go for it. As long as tests pass, we know we're not hitting the problematic code path, as uneasy as it makes me feel.

PR #1493, opened 2022-10-08
Graphql endpoint

GavinMendelGleason:
Implement a GraphQL endpoint for a data product.

Includes
* GraphQL Schema creation from Terminus Schema
* GraphQL queries automatically constructed from the TerminusSchema

Not all GraphQL or TerminusDB features will be implemented in this first pass, but it should work decently for what is exposed on  arbitrary TerminusDB data products.


rrooij (approved):
Nice job!

PR #1533, opened 2022-10-25
various tweaks to the openapi spec for the document interface

matko:
I noticed the openapi spec for the document interface contained some errors and could be a bit more complete. It's still not all there, but hopefully this is an improvement.


docs/openapi.yaml
```diff
@@ -456,7 +456,7 @@ paths:
           description: Specific document ID to look for
           schema:
             type: string
-            example: Person_Robin_1991-02-01
+            example: Person/Robin_1991-02-01
```
rrooij, line 459: Thank you for making me a proper Person type

rrooij (approved):

PR #1542, opened 2022-11-01
Dockerfile: use patched version of swipl

rrooij:
Even though the segfault does not seem to happen on the Debian stable versions, I still feel very uncomfortable with this not patched in.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ARG DIST=community
 
 # Install the SWI-Prolog pack dependencies.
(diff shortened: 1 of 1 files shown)


rrooij:
For some extra context, the changes between the images can be found here:

https://github.com/terminusdb/swi-prolog-docker/commit/c2dec6ade7a05a4c3b13e546bea5d0f6d972f5d1

matko (approved):
Even if this particular bug is not hitting us right now on this particular container version of debian, I like that you've set up the infra for patch application cause I think we're gonna need it several times more in the future.

PR #1555, opened 2022-11-10
snapcraft: add m4 as dependency

rrooij:
It was missing m4 as a dependency for the build of the GMP rust crates.


rrooij:
Not needed anymore since #1557 solved this issue.

Issue #1560, opened 2022-11-12
Inherited properties behave differently from local properties


hoijnet:
**Describe the bug**
Set properties behave differently if property is inherited. In the example below, the type `myInheritedType` inherits the property `inheritedSetString` from a superClass (Set of strings). The localSetString definition is a Set of strings, where the property is applied to the local class.

When saving the document, TerminusDB treats the two string set properties differently. In the past I'm quite sure that both behaved as `localString` (presenting the one item as an array), as I have not had any issues managing previous data. With the inconsistent handling, it breaks editing in my tool, and is inconsistent and probably breaks for others too.

This also happens for Link sets, have not tested with other types of Sets.

```
{
  "@id": "myInheritedType/e63b267d08c547302e6beb10bb2b15c0a1d9ca037659d60b5dffc77a0197a869",
  "@type": "myInheritedType",
  "inheritedSetString": "myString",
  "localSetString": [
    "myString"
  ]
}
```

**To Reproduce**
1. Create one type `myType` with a Set string property `inheritedSetString`
2. Created a subclass `myInheritedType` inheriting `myType` with a String set property `localSetString`
3. Create a document of type `myInheritedType`
4. The document looks like below which is inconsistent.

```
{
  "@id": "myInheritedType/e63b267d08c547302e6beb10bb2b15c0a1d9ca037659d60b5dffc77a0197a869",
  "@type": "myInheritedType",
  "inheritedSetString": "myString",
  "localSetString": [
    "myString"
  ]
}
```

**Expected behavior**
I would expect above document to look like this:

```
{
  "@id": "myInheritedType/e63b267d08c547302e6beb10bb2b15c0a1d9ca037659d60b5dffc77a0197a869",
  "@type": "myInheritedType",
  "inheritedSetString": [
    "myString"
  ]
  "localSetString": [
    "myString"
  ]
}
```

Note that oneItemWillBePresentedAsString is not an array (which is inherited from myType).

**Info (please complete the following information):**
 - Tested with 10.1.7 running in Docker
 - Tested with 10.1.7 running in TerminusX

**Additional context**

GavinMendelGleason:
We'll take a look and get it fixed ASAP.

hoijnet:
It works again now in the 10.1.9 release, thanks a lot!! Confirmed in the docker image.

🙏

PR #1567, opened 2022-11-15
Fix set property inheritance in doc retrieval

matko:
we accidentally walked the inheritance graph in the wrong direction.


GavinMendelGleason:
Fixes #1560

rrooij (approved):

PR #1575, opened 2022-11-18
Graphiql no fixed host

rrooij:
The graphiql endpoint always assumed `terminusdb` to be accessible on localhost. Mostly because the `graphiql` endpoint is intended as a tool for developers. However, there are certainly cases where you would like to access it under different domains.

Therefore I removed the host and port from the path as it now assumes that if you visit for instance:

`http://some-site.xyz:9090/graphiql/admin/test` that your terminusdb server is also on `http://some-site.xyz:9090`


matko (approved):

matko:
I'll merge this, but nevertheless I need to point out that the graphiql endpoint was never really meant for production use. It's only a stopgap solution until we embed graphiql properly into the dashboard.

PR #1576, opened 2022-11-21
Fix bug reported on GraphQL id filtering

GavinMendelGleason:
Adds type filtering to the explicit ID lookup argument, which avoids having erroneous results for objects of a different type.


GavinMendelGleason:
Already in GraphQL backlinks branch

PR #1602, opened 2022-12-06
server: add JSON formatted welcome message

rrooij:
If the LOG_FORMAT ENV variable is set to json, it will print a JSON welcome message instead of the normal text welcome message.

This is better because it is awkward if the logs are streamed to a JSON parser but it does not always contain valid JSON.

The welcome message in this particular commit looks like this:

```
{
  "message": "Welcome to TerminusDB ! You can view your server in a browser at http://localhost:6363",
  "version": "10.1.11",
  "args": "[serve,help(false),interactive(false),memory(_250)]"
}
```

### DIFF
--- a/src/server/main.pl
+++ b/src/server/main.pl
@@ -20,6 +20,7 @@
                                        jwt_jwks_endpoint/1,
                                        server/1,
                                        server_port/1,
+                                       log_format/1,
                                        worker_amount/1,
                                        is_enterprise/0,
                                        terminusdb_version/1]).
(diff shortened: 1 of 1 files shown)


src/server/main.pl
```diff
@@ -116,15 +117,24 @@
         p('TerminusDB is still synchronizing backing store')
     ]).
 
+
```
GavinMendelGleason, line 120: This needs to either add a cut after json log format, or use picat syntax with a head constraint.

GavinMendelGleason (changes requested):

src/server/main.pl
rrooij: Done! Thanks

GavinMendelGleason (approved):

PR #1622, opened 2022-12-23
Adding graphql inheritance test and implementation

GavinMendelGleason:

### DIFF
--- a/src/rust/terminusdb-community/src/graphql/frame.rs
+++ b/src/rust/terminusdb-community/src/graphql/frame.rs
@@ -679,7 +679,6 @@ impl AllFrames {
                 }
             }
         }
-        eprintln!("{subsumption_rel:?}");
         self.subsumption = Some(subsumption_rel);
     }
 }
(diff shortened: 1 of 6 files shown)


src/rust/terminusdb-community/src/graphql/frame.rs
```diff
@@ -636,6 +651,37 @@ impl AllFrames {
...
+                for superclass in supers {
+                    if let Some(v) = subsumption_rel.get_mut(&superclass) {
+                        v.push(class.to_string());
+                    } else {
+                        subsumption_rel.insert(superclass, vec![class.to_string()]);
+                    }
+                }
+            }
+        }
+        eprintln!("{subsumption_rel:?}");
```
matko, line 682: print should go

src/rust/terminusdb-community/src/graphql/mod.rs
```diff
@@ -39,6 +36,7 @@ predicates! {
         let frames: AllFrames = context.deserialize_from_term(&frame_term).expect("Unable to parse frames into rust struct");
         let mut sanitized_frames: AllFrames = frames.sanitize();
         sanitized_frames.invert();
+        sanitized_frames.calculate_subsumption();
```
matko, line 39: I think we should deserialize frames to an intermediate type, then convert that to our final type to do these finalization steps you have in here. That way it's more clear that these steps only happen once as part of a construction, instead of being methods that can be called at arbitrary times.

matko (changes requested):
just remove the eprintln and it is all good

matko (approved):

PR #1632, opened 2023-01-03
fix base class not being included in subsumed document retrieval

### LINKED ISSUE #1614: Document interface returns only subsumed documents for top level types
**Describe the bug**
Document interface returns only subsumed documents for top level types, both in TerminusX and through the document interfaces. See below to reproduce.

**To Reproduce**
1. Create a new data product in TerminusX
2. Create a top-level document type `Entity` with optional string label property `label`
3. Create a document of type `Entity` (note that the document can be listed in the document_explorer)
4. Create a subsumed document type `Statement` that inherits `Entity`
5. Create a document of type `Statement` (verify that it can be listed under `Statement` in document_explorer)
6. Go to document_explorer root, see that there is 1 document for `Entity`, and 1 document for `Statement`
7. Click on `Entity` documents, only the `Statement/...` document is visible
8. Delete the `Statement/...` document
9. Note that the `Entity` document can't be listed (it is not retrieved from the document interface)
10. Delete the `Statement` type
11. View that the `Entity/...` document can be listed again.

**Expected behavior**
The `Entity` documents should be retrieved from the document interface whether or not a subsumed document type has been created. When listing documents of type `Entity`, both `Entity` and `Statement` documents should be listed; not only `Statement` documents.

**Screenshots**
image

image

Note that 4 documents are created in total. 1 out of 3 are shown for `Entity`... Which is completely incorrect: 4 should be shown,  4 is the total. What is shown is 1 of 3.

**Info (please complete the following information):**
 - OS: TerminusX dashboard from Edge on macOS
 - Same behaviour when using DFRNT as client.

**Additional context**
It seems there is something odd going on in the document interface.

matko:
Fixes #1614


GavinMendelGleason (approved):

hoijnet:
Just confirming back this is resolved in 10.1.12 (in DFRNT), but not yet in TerminusX.

PR #1678, opened 2023-01-25
V11.0.0 release notes

GavinMendelGleason:


matko (approved):

matko:
This also upgrades store to 0.20.0. Kinda weird that this is the PR that does it but it does need to happen anyway.

PR #1685, opened 2023-01-30
Use patched juniper with updated smartstring dependency

matko:
It turns out that the version of smartstring that juniper is using has some weird issues which caused certain strings to get mangled unpredictably. In order to quickly resolve this, we've set up a temporary juniper fork where we've upgraded the smartstring dependency to version 1.0, which doesn't appear to have the same issue.

This is obviously a temporary fix to resolve a blocker for us. We should coordinate with the juniper maintainers to hopefully get a proper fix of juniper out, or alternatively wait for their release of 0.16 (which seems to involve a lot of changes to their codegen, so unfortunate;y that won't be an easy upgrade).

### DIFF
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -36,12 +36,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
-version = "1.0.68"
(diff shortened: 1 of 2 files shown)


matko:
#1686

PR #1695, opened 2023-01-31
limit the default password hash cost to 10 for less roundtrips

matko:
It turns out that password hashing is taking up a significant amount of time in TerminusDB due to the hash function being set to cycle 2^17 times, leading to hash costs in excess of 100ms. This is completely unacceptable for an API.

This PR limits the cost to 10, which means 1024 cycles. IMO there's little need for any cycling so I'm open to lowering this even more. At this point though it looks like costs for single requests fall into the noise.

### DIFF
--- a/src/config/terminus_config.pl
+++ b/src/config/terminus_config.pl
@@ -31,7 +31,8 @@
               pinned_databases/1,
               plugin_path/1,
               dashboard_enabled/0,
-              parallelize_enabled/0
+              parallelize_enabled/0,
+              crypto_password_cost/1
           ]).
 
 :- use_module(library(pcre)).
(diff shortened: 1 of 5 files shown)


matko:
Note that this change just affects passwords generated after this change. Existing passwords should still validate (and take a long time in doing so), but people can easily achieve faster speeds just by changing their password, and new users and fresh stores will start out on those faster speeds.

GavinMendelGleason (approved):

matko:
As a rough timing indication from my local machine:
With cost set to 0, doing 1000 calls to api/ok with basic auth takes 0m8.349s.
With cost set to 10, doing 1000 calls to api/ok with basic auth takes 0m9.388s.

PR #1699, opened 2023-02-03
Schema migration

GavinMendelGleason:
This is a work in progress representing the first steps in schema migration.  The steps which will be performed in the database are as follows:

1. We produce a command language capable of performing schema updates, and which can be compiled to produce instance updates consistently, and able to migrate between *most* schemata. 
2. We interpret the command language to create a before and after schema
3. We compile the command language to produce a number of instance update commands
4. We interpret the instance updates to produce an updated instance graph, paste in the new schema atomically, and commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
 
 env:
-  LAST_RELEASED_VERSION: 11.0.2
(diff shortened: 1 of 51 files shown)


matko:
Need a way to migrate collections (set to list/array etc)

matko (approved):

PR #1708, opened 2023-02-08
Adding reorder of capabilities, and make anonymous work as fallback

GavinMendelGleason:

### DIFF
--- a/src/core/account/capabilities.pl
+++ b/src/core/account/capabilities.pl
@@ -121,19 +121,20 @@ auth_action_scope(DB, Auth, Action, Scope_Iri) :-
         (
             t(Auth, capability, Capability),
             t(Capability, role, Role),
+            t(Role, action, Action)
             t(Capability, scope, Intermediate_Scope_Iri),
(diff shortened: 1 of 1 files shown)


src/core/account/capabilities.pl
```diff
@@ -121,19 +121,20 @@
         (
             t(Auth, capability, Capability),
             t(Capability, role, Role),
+            t(Role, action, Action)
```
matko: comma at end

src/core/account/capabilities.pl
```diff
@@ -121,19 +121,20 @@
...
             t(Capability, scope, Intermediate_Scope_Iri),
             path(Intermediate_Scope_Iri, (star((p(child);p(database)))), Scope_Iri, _),
```
matko: shouldn't have comma here

matko (approved):

src/core/account/capabilities.pl
GavinMendelGleason: Fixed

PR #1709, opened 2023-02-08
allow impersonation of users in the cli

matko:
The CLI does all its actions as admin. It's convenient to be able to perform them as another user too, so that authorization checks run. This PR introduces an `--impersonate` flag which can be used to pass in the user that is performing the action.


docs/terminusdb.1.ronn
```diff
@@ -27,838 +27,14 @@ Display help regarding terminusdb.
 
 ### test
 
-`terminusdb test OPTIONS`
```
rrooij, line 30: What happened to all the remove docs in the manpages?

rrooij (dismissed):
Something weird went on with the manpage

docs/terminusdb.1.ronn
matko: oh that's interesting. This seems to be an issue with the automatic manpage generation. I'm not sure what's going on there.

matko: It's all coming from this automated commit: https://github.com/terminusdb/terminusdb/pull/1709/commits/a11d4aa7f54be8abf90412b439cdf4011d3ec182

rrooij: I see why it doesn't work. Your changes introduced a regression in markdown generation of the help page.

Try `./terminusdb help --markdown` in your branch vs `main`.

matko: Looks like the markdown generation wasn't able to handle a flag that didn't have a short version.
I hope I just fixed that.

matko: yup, manpage is back.

PR #1719, opened 2023-02-10
Support insertion of nested top-level documents with key dependencies

matko:
This PR introduces support for insertion of nested documents with key dependencies.

This was previously not working because we wouldn't know the id in time to do this, as id generation of parents strictly happened before id generation of children.

This PR reorders the id generation, so that key fields have their id generated first.

### DIFF
--- a/src/core/document/json.pl
+++ b/src/core/document/json.pl
@@ -10965,6 +10965,52 @@ test(normalizable_float,
            foo: "0.5000000"},
         'Thing/0.5').
 
+nested_document_schema('
+{ "@base": "terminusdb:///data/",
+  "@schema": "terminusdb:///schema#",
+  "@type": "@context"}
+
+{ "@type" : "Class",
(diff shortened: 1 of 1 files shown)


matko:
There is an issue here.
It is apparently possible to have a key dependency to a subdocument too. I'm not sure how that was ever supposed to work but the schema checker lets it through.

matko:
> There is an issue here. It is apparently possible to have a key dependency to a subdocument too. I'm not sure how that was ever supposed to work but the schema checker lets it through.

Note that this has never worked properly. The current situation results in an ugly stack trace (the one we're trying to fix for the 'document' case). But unlike the document case, we cannot fix this. This is a cyclical dependency.

matko:
Marking as ready for review cause the other bug is orthogonal. It needs to be fixed but this is no less wrong.

GavinMendelGleason (approved):

PR #1749, opened 2023-03-03
Fix graphql ne queries

GavinMendelGleason:


src/rust/terminusdb-community/src/graphql/query.rs
```diff
@@ -120,6 +120,9 @@ FilterObject{ edges:
 fn compile_string_input_value(string_type: &str, value: StringFilterInputObject) -> FilterType {
     if let Some(val) = value.eq {
         FilterType::String(GenericOperation::Eq, val, string_type.to_string())
+    } else if let Some(val) = value.ne {
+        eprintln!("value is ne");
```
matko, line 124: probably shouldn't print here

matko (changes requested):

matko (approved):

PR #1756, opened 2023-03-08
Adding Patch Endpoint for Resources

GavinMendelGleason:
This endpoint will apply a patch to a resource

### DIFF
--- a/src/core/api/api_patch.pl
+++ b/src/core/api/api_patch.pl
@@ -19,6 +19,7 @@
 :- use_module(library(lists)).
 :- use_module(library(plunit)).
 :- use_module(library(option)).
+:- use_module(library(assoc)).
 
 api_patch(_System_DB, _Auth, Patch, Before, After, Options) :-
     % no auth yet.
--- a/src/core/api/api_patch.pl
+++ b/src/core/api/api_patch.pl
@@ -1,6 +1,6 @@
 :- module(api_patch, [
+              api_patch_resource/5,
               api_patch/6,
-              api_patch/7,
               api_diff/6,
               api_diff_id/8,
               api_diff_id_document/8,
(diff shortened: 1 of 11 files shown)


src/server/routes.pl
```diff
@@ -2488,6 +2492,36 @@
...
+            _{ patch : Patch } :< JSON),
+        error(bad_api_document(JSON, [patch]), _)),
+    % We should take options about final state here.
+    api_report_errors(
+        patch,
+        Request,
+        (   (   memberchk(search(Search), Request)
+            ->  true
+            ;   Search = []),
+            param_value_search_or_json_required(Search, JSON, author, text, Author),
```
matko, line 2512: There is a weird inconsistency here between endpoints.
Some endpoints seem to expect commit info to be in a special commit_info subdocument in the payload. But this (and I presume other places where this was copied from) would expect author and message to be toplevel.

src/core/api/api_patch.pl
```diff
@@ -12,16 +13,69 @@
...
+        (
+            patch_id_pairs(Patch, Patch_And_Ids),
+            mapm({Context, Merged_Options}/[Patch-_,Conflict,Ids]>>(
+                     apply_diff_ids_captures(Context, Patch, Conflict, Ids, Merged_Options)
+                 ),
+                 Patch_And_Ids,
+                 Conflicts,
+                 Ids_List,
+                 Empty,
+                 _Captures
```
matko, line 62: if we thread through captures shouldn't we be checking to make sure that all captures are bound in the end?

Unless of course there is no way to actually use captures and this was just to make the calls happy.

src/core/api/api_patch.pl
```diff
@@ -12,16 +13,69 @@
...
+                ),
+            % actually an unzip
+            exclude([null]>>true, Conflicts, Witnesses),
+            (   Witnesses = []
+            ->  true
+            ;   maplist([conflict(Conflict),Conflict]>>true,
```
matko, line 68: should this not be an include?
This will only succeed if all patch fragments had conflicts.

GavinMendelGleason: Yes, this is wrong.

src/core/api/api_patch.pl
```diff
@@ -12,16 +13,69 @@
...
+                        Witnesses,
+                        Conflicts),
```
matko, line 70: this rebinds Conflicts, probably not great.

matko (changes requested):

src/server/routes.pl
GavinMendelGleason: Let's fix that where the inconsistency arises.

matko (approved):

PR #1914, opened 2023-08-02
vscode project files

matko:
This PR introduces vscode project files for those that want to use vscode.

Specifically, this PR introduces a bunch of launch, build and test tasks, and common workspace settings, as well as some scripts needed to set up an environment where the various targets can find the right version of SWI-Prolog.
This setup requires cargo-swipl to be installed.

## launch tasks
- debug terminusdb server in LLDB
- debug a swipl repl through LLDB with terminusdb loaded, but no server started

## build tasks
- build the rust module
- build terminusdb

## test tasks
- run rust unit tests through cargo test
- run the terminusdb prolog tests

## common settings
- discovery of nested rust project
- ensure that lldb can find the swi-prolog library
- ensure that rust-analyzer uses the `cargo-swipl` wrapper to run all cargo commands
- ensure that launched terminals have swi-prolog environments set

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,43 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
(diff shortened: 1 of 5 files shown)


rrooij (approved):
Awesome. Not sure why VSCode accepts comments in JSON, but if it allows this then it's OK.

PR #1917, opened 2023-08-03
New parameter 'ids' on document get endpoint to retrieve multiple documents by id at once

### LINKED ISSUE #1226: A way to get several documents at once by submitting a list of ids
Currently the document interface takes an id field, but it would be convenient to be able to supply a list of ids instead.

matko:
Fixes #1226.

### DIFF
--- a/src/cli/main.pl
+++ b/src/cli/main.pl
@@ -1077,6 +1077,12 @@ opt_spec(doc,get,'terminusdb doc get DATABASE_SPEC OPTIONS',
            shortflags([i]),
            default('_'),
            help('id of document to retrieve')],
+          [opt(ids),
+           % todo something something comma separated values
+           type(term),
+           longflags([ids]),
+           default('_'),
+           help('list of document ids to retrieve')],
           [opt(type),
            type(atom),
            longflags([type]),
@@ -2159,9 +2165,14 @@ run_command(doc,get, [Path], Opts) :-
     option(as_list(As_List), Opts),
     option(unfold(Unfold), Opts),
     option(id(Id), Opts),
+    option(ids(Ids_Param), Opts),
     option(type(Type), Opts),
     option(compress_ids(Compress_Ids), Opts),
     option(query(Query_Atom), Opts),
+
+    (   ground(Id)
+    ->  Ids = [Id]
+    ;   Ids = Ids_Param),
 
     (   var(Query_Atom)
     ->  Query = Query_Atom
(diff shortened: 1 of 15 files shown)


matko:
I rewrote the document endpoint so that getting by a single id is now a special case of getting by several ids. I then created new parameters in both the cli tool and the server endpoint to use this with multiple ids (supporting both as a query param or inside of a submitted query document). If a single id is submitted, this will take precedence over any id list submitted.

hoijnet:
This is so beautiful!!!

src/server/routes.pl
```diff
@@ -497,9 +505,9 @@
 
             api_read_document_selector(
                 System_DB, Auth, Path, Graph_Type,
-                Id, Type, Query, Config,
+                Id, Ids, Type, Query, Config,
                 Requested_Data_Version, Actual_Data_Version,
-                cors_json_stream_write_headers_(Request, Actual_Data_Version)
+                routes:cors_json_stream_write_headers_(Request, Actual_Data_Version)
```
rrooij, line 510: ??

src/cli/main.pl
```diff
@@ -1077,6 +1077,12 @@
            shortflags([i]),
            default('_'),
            help('id of document to retrieve')],
+          [opt(ids),
+           % todo something something comma separated values
```
rrooij, line 1081: ? :O

rrooij (approved):

src/server/routes.pl
matko: yeah i don't know. it started complaining here. slightly cargo cult but what can you do.

src/cli/main.pl
matko: the cli now requires you to pass in ids like so:
./terminusdb doc get admin/foo --ids '["Foo/1","Foo/2"]'

So as a json list.

Would probably be nicer if you could directly specify a comma separated list.

matko:
> If I expect IDs to be returned, I want an error. If I ask exploratively I want what I can get. In the my current use case I would expect an error is the ID can't be found, as it implies that a new revision was likely created between the time of check and the time of update/request of my model.

Specifically on this point, note that we actually support data version headers which assert that the data you're querying is at the revision you expect it to be.

Every request that queries or updates the database will result in a `TerminusDB-Data-Version` header in the response. In any subsequent query/update request, this header can be resubmitted, and the operation will fail if the version mismatches.

PR #1918, opened 2023-08-03
snap: copy src folder in final snap

rrooij:
We rely on some files that are present in the src folder. For instance, the ref.json. We remove the rust folder though because the artifacts are really big and we don't rely on those files at runtime.


dfrnt-SigmundFuchs:

PR #1953, opened 2023-09-01
Fix compression of @base in document interface

GavinMendelGleason:
If there was a prefix used which compressed with the same URL as `@base`, then this prefix was used preferentially. This has been changed so that the `@base` prefix is used preferentially.


dfrnt-SigmundFuchs:

PR #1966, opened 2023-09-25
properly detect cycles in unfoldables when collections are involved

matko:
This ensures the schema checker will reject schemas where collections of unfoldables lead to cycles.


matko:
Closed in favor of #1967 which moves the cycle checking to runtime.

PR #1967, opened 2023-09-25
untrusting cycle detection during document extraction

matko:
This changes the document extraction logic so that unfold cycles are expected and handled by emitting an IRI instead of an unfolded document.

### DIFF
--- a/src/core/document/json.pl
+++ b/src/core/document/json.pl
@@ -6333,7 +6333,7 @@ test(bad_documentation,
         _
     ).
 
-test(bad_unfoldable,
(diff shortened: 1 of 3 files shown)


matko:
The reason to call this untrusting is that it doesn't trust any document extraction. In practice though, due to our schemas, we know a lot of cycle situations are plain impossible and there's little point in checking.
Since we just want this for the unfoldable case, it might make sense to only check when traversing a link to an unfoldable document.

GavinMendelGleason (approved):

PR #1985, opened 2023-10-13
retrieve and query foreigns from graphql

matko:


matko:
Closing in favor of #1971 which has this work merged.

PR #2018, opened 2023-11-01
Db list by organization

matko:
This introduces a new endpoint, `api/db/<org>` which will list all databases accessible by the authenticated user under the specified organization.

In addition, this PR contains various optimizations to db listing itself.


GavinMendelGleason (approved):
Looks good - hopefully the test coverage for listing is already good enough? Might be good to test the new behaviour in the js integration tests to avoid regressions.

matko:
I'll add some integration tests before release.

PR #2041, opened 2023-11-24
Delete documents by type

matko:
This PR introduces document deletion by type.
It also upgrades to the latest version of store.


GavinMendelGleason (approved):
Excellent

PR #2070, opened 2023-12-24
SSL for TerminusDB with Traefik and Let’s Encrypt

alexander-mart:
## Done

- [x] terminus-db dashboard
- [x] change-request-api

## To improve

- [ ] enable vectorlink:

```sh
docker logs 9efd97ccffa9
time="2023-12-24T21:12:32Z" level=info msg="Configuration loaded from flags."
time="2023-12-24T21:12:32Z" level=error msg="service \"vectorlink-terminusdb\" error: port is missing" providerName=docker container=vectorlink-terminusdb-a80b620d2d5ef46c35a1035b5c8f64f22628c9420fafe67ceaa8dff89f757224
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,51 @@
+DOMAIN=localhost
+
+LETSENCRYPT_EMAIL=admin@localhost
+
+# WARNING! Note that you should leave `CERT_RESOLVER` variable empty if you test your deployment locally. The password is `admin` and you might want to change it before deploying to production.
+CERT_RESOLVER=
+
+# Traefik dashboard basic auth login:
+TRAEFIK_USER=admin
+
+# Generate hash of traefik dashboard basic auth password with:
+# $ htpasswd -nBC 10 admin
+# 
+# Or use online version: https://www.web2generators.com/apache-tools/htpasswd-generator
+# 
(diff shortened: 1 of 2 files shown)


matko:
Unfortunately, this change would break vectorlink communication as right now it relies on being able to use insecure authentication.

I created an issue in vectorlink: https://github.com/terminusdb-labs/vectorlink/issues/21

alexander-mart:
@matko If I don't need `vectorlink`, is this configuration safe to use in production? Do you think everything else is ok?

matko:
> @matko If I don't need `vectorlink`, is this configuration safe to use in production? Do you think everything else is ok?

I'm not sufficiently experienced with traefik to comment on its configuration, but focusing purely on the terminusdb part, commenting out the  lines doing the insecure header forward should be sufficient to make this safe to expose online.

hoijnet:
I believe the Traefik dependency should not be in the main docker-compose configuration file as there are many ways to run it.

I would suggest to create a folder distribution/docker-compose-ssl with the docker-compose.yml file, so that others can use your example.

Please update the PR for a new review.

dfrnt-SigmundFuchs:

PR #2099, opened 2024-02-20
make document API predicates interruptible

matko:


matko:
This should probably be more conservative about calling handle_signals. It's hard to find a good balance between interrruptibility and throughput, and this problem is likely to keep coming up in other long-running predicates.
I'm keeping this a draft while working out another approach, probably something budget driven with a counter.

dfrnt-SigmundFuchs:

PR #1945, opened 2023-08-16
ci: don't depend on external infra

rrooij:
It is better for the community if it is only depending on GitHub infra instead of custom AWS infrastructure or BuildJet machines.

PR #2132, opened 2024-08-07
disabled self-hosted ci runner

matko:
pretty much #1945 but we're keeping the buildjet arm runners.


hoijnet:
Self-hosted runner is up and running again. Closing this PR again.

PR #2151, opened 2025-03-28
Fix docker build

GavinMendelGleason:

### DIFF
--- a/.github/workflows/arm.yml
+++ b/.github/workflows/arm.yml
@@ -33,7 +33,7 @@ jobs:
         run: gzip terminusdb-server-docker-image-arm64.tar
 
       - name: Upload Docker image
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         with:
             name: terminusdb-server-docker-image-arm64
             path: terminusdb-server-docker-image-arm64.tar.gz
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,11 +7,10 @@ jobs:
 
   build:
     name: Docker image
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
 
     steps:
       - uses: actions/checkout@v3
-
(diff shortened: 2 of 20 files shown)


hoijnet (approved):

hoijnet:
We still have an issue with a rust error, unsure if we saw it last time?

I get the same result locally as with the now-put-in-place self hosted runner. @GavinMendelGleason, I trusted the compiler on this one. I

```
#26 45.89 error[E0283]: type annotations needed
#26 45.89   --> terminusdb-community/src/graphql/frame.rs:38:35
#26 45.89    |
#26 45.89 38 |         self.0.partial_cmp(&other.into())
#26 45.89    |                                   ^^^^
#26 45.89    |
#26 45.89    = note: multiple `impl`s satisfying `_: From<&str>` found in the following crates: `bytes`, `hyper`:
#26 45.89            - impl From<&'static str> for bytes::bytes::Bytes;
#26 45.89            - impl From<&'static str> for hyper::body::body::Body;
#26 45.89    = note: required for `&str` to implement `std::convert::Into<_>`
#26 45.89 help: try using a fully qualified path to specify the expected types
#26 45.89    |
#26 45.89 38 -         self.0.partial_cmp(&other.into())
#26 45.89 38 +         self.0.partial_cmp(&<&str as std::convert::Into<T>>::into(other))
#26 45.89    |
```

Unsure if the change is correct, it is based on the compiler suggestion and the struct signature, changed it to this in frame.rs:

```
self.0.partial_cmp(&<&str as Into<std::borrow::Cow<'_, str>>>::into(other))
```

```
#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Deserialize, Default, Hash)]
pub struct GraphQLName<'a>(pub Cow<'a, str>);
```

It now compiles and I think that we should find any issues in the tests, but also it looks to be correctly annotated now.

Issue #2161, opened 2025-04-09
There is an error in test_utils.pl that prevents certain integration tests from running


hoijnet:
**Describe the bug**
Integration tests are disabled currently due to issue in read_line_until_start_line() of test_utils.pl that yields server_has_no_output in connection with process_create(), likely due to either timeouts or a string matching failure in the server, and should be fixed in upcoming releases.

**To Reproduce**
1. Open routes.pl
2. See the commented out tests
3. Uncomment the tests
4. Build TerminusDB: `docker build -t terminusdb .`
5. Run unit tests: `docker run -it terminusdb make test`
6. See output

**Expected behavior**
All unit tests should pass, including the more integration test oriented ones

**Screenshots**

Example of an error:

```
% PL-Unit: clone_endpoint 
ERROR: /app/terminusdb/src/server/routes.pl:845:
        error in setup: Unknown error term: server_has_no_output
```

hoijnet:
Unsure if it should be a separate error, if so, split them. On the self-hosted runner, the terminusdb is left lying around after the integration tests are completed. It's likely we need to add a cleanup in the build.yml or check.yml script.

GavinMendelGleason:
This will likely affect any of the route tests as they all start a server to test the endpoints.

hoijnet:
> This will likely affect any of the route tests as they all start a server to test the endpoints.

Actually it was only a few of them that were affected. Many of them still work, I think they use a different mechanism!

Let's work on this one in #2162 for the next release. I will change the uncommenting to the blocked() in that PR.

hoijnet:
> I will change the uncommenting to the blocked() in that PR.

The PR there contained the blocked statements instead and it's clear from looking at the tests what is blocked and that they need addressing!

dfrnt-SigmundFuchs:
@GavinMendelGleason @hoijnet These tests fail because `https://cdn.terminusdb.com/jwks.json` cannot be resolved.

```prolog
[server stdout]   [46] throw(error(existence_error(url,'https://cdn.terminusdb.com/jwks.json'),context(_92342,...)))
[server stdout]   [44] http_open:try_http_proxy(direct,[uri('https://cdn.terminusdb.com/jwks.json'),...|...],_92384,'<garbage_collected>') at /usr/lib/swipl/library/ext/http/http/http_open.pl:512
[...]

PR #2162, opened 2025-04-09
Issue/upgrade to swipl 9.2.9

hoijnet:
This PR enables us to upgrade to swipl 9.2.9:
* Updates in swipl [9.0.3 --> 9.2.9](https://www.swi-prolog.org/ChangeLog?branch=stable&from=9.0.3&to=9.2.9)
* Performance update on some platforms with arithmetic
* Additional bug fixes and enhancements
* Keeping it separate from other code

Suggestion is to get issue #2161 fixed and the integration tests running again in the next release and not commit this PR until the integration tests are running again too with 9.2.9.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -36,7 +36,7 @@ jobs:
 
     steps:
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
 
       - name: Download Docker image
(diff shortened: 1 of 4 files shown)


hoijnet:
To complete this PR, we should ensure that the build scripts are updated first to ensure the self-hosted running is removed again in time so that future integration tests continue to run.

dfrnt-SigmundFuchs (approved):

PR #2166, opened 2025-04-09
Steps 3,4,5,6: Updated version to v11.1.14 and next to v11.1.15

hoijnet:
This completes the release for v11.1.14 once all has been built, approved and merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
 
 env:
-  LAST_RELEASED_VERSION: 11.1.13
-  CURRENT_REPO_VERSION: 11.1.14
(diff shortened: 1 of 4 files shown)


hoijnet:
@dfrnt-SigmundFuchs and @GavinMendelGleason , asking Gavin to approve this one to ensure we followed the process correctly. Let's identify how we want to do the next release!

If all goes well, my understanding is that we should have a new docker image on docker hub once this PR completes.

hoijnet:
The image is already there, so getting this PR through is more about closing the books. Thus requesting the review from @dfrnt-SigmundFuchs  should be all right! Seems all things are working correctly!

GavinMendelGleason (approved):

PR #2191, opened 2025-04-27
Don't use system gmp on macOS non-docker builds

hoijnet:
Homebrew installs gmp in a non-system location and is not installed on systems in general. So should not depend on system libraries on macOS


.github/workflows/non-docker.yml
```diff
@@ -47,6 +47,10 @@ jobs:
         if: runner.os == 'macOS'
         run: brew install protobuf
 
+      - name: Use compile gmp in on macOS
+        if: runner.os == 'macOS'
+        run: sed 's/gmp-mpfr-sys = {version = "~1.5", default-features = false, features=\["use-system-libs"\]}/gmp-mpfr-sys = {version = "~1.5", default-features = false, features=\[\]}/' ./src/rust/terminusdb-community/Cargo.toml
```
dfrnt-SigmundFuchs, line 52: This currently only prints the filtered result. It would be better to parametrize the make build process properly. e.g. 

Makefile
```
FEATURES ?= ""

build:
	cargo build --no-default-features --features "$(FEATURES)"
```
 and then
```
- name: Build without system libs
  run: make build FEATURES=""
```

dfrnt-SigmundFuchs (changes requested):

dfrnt-SigmundFuchs:
@hoijnet `non-docker-os.yml` sounds cumbersome; it's simply a native build, or just a build. Let's rename it accordingly.

.github/workflows/non-docker.yml
dfrnt-hoijnet: The idea was to have `sed -i`, but forgot to change, good catch and saves us a cycle! Also much better with the suggestion you made to make it generic. This was the simple solution, yours the good solution!

Would you be up for making the changes necessary?

Now we know how to fix the macOS build I think, at least this one, and we have the name change and also the Linux devel version left.

Probably a note could be added in the root folder in a troubleshooting guide regarding GMP. Unsure why Rust complaint regarding the Linux build there but maybe a later version of some libraries or later rust?

I believe few will want to use native system libs on MACOS, whereas on Linux more will want to.

dfrnt-hoijnet:
> @hoijnet `non-docker-os.yml` sounds cumbersome; it's simply a native build, or just a build. Let's rename it accordingly.

Agreed, thinking this could be an alternative:
nightly-build-verification.yml

dfrnt-SigmundFuchs:
> Agreed, thinking this could be an alternative:
nightly-build-verification.yml

Maybe we should reverse the approach: build the Docker image on a schedule (if changes are detected, on manual request, and before a release), for example once a day — although once a week might be sufficient. For every PR, we should run a standard build to check that the code compiles and passes tests.

see #2194

dfrnt-SigmundFuchs (approved):

PR #2201, opened 2025-05-04
Issue/2175 path missing delimiter

hoijnet:
This PR first corrects the incorrect unit test for metadata, where the setup was incorrectly made, hiding the bug for the size WOQL predicate. Now the metadata unit tests correctly gives below error, matching production use. 

```
% PL-Unit: metadata 
ERROR: /app/terminusdb/src/core/query/metadata.pl:95:
        test count_and_size: received error: size_file/2: file `'/tmp/temporary_terminus_store/1albk2r9czxblb7vzutngvavz3xjeny409/4091b174a9e435d6b3bdb14d62fc615df85cbe79.larch'' does not exist (No such file or directory)
make: *** [Makefile:48: test] Error 1
``` 

Next, the bug is fixed, making it possible to use the size WOQL predicate, where the missing delimiter is added, fulfilling the metadata test. Tested locally with WOQL on a newly built TerminusDB, with: `size("admin/task-set-local/local/branch/main", "v:out")` for the admin organization and the task-set-local data product. It now correctly gives the size of the data in the data product.


dfrnt-SigmundFuchs:
@hoijnet If this is a fix, do we have a test case that proves it's fixed - i.e., a test expected to fail?

hoijnet:
@dfrnt-SigmundFuchs, the reason the test did not catch it in the past was that the test was wrong. This is what I did, which is in essence working backwards, I'll explain why after the list.

1. Confirm the issue in a build
2. Fix the issue
3. Confirm the issue was fixed in a new build
4. Fix the test and revert the fix
5. Confirm that the test reported the same issue I got in step 2
6. Fix the issue again
7. Confirm the issue was fixed in a new build, and that the new build passes the test.

The issue was that the test setup was improperly made and did not detect the issue (due to the slash being present in the test preparation, which is not what is normally processed, without the slash.)

Thus, I have confidence in it being confirmed working and that the test was wrong.

dfrnt-SigmundFuchs (approved):

PR #2205, opened 2025-05-05
Reorg docker build actions (1/2)

dfrnt-SigmundFuchs:
This PR extracts Docker image builds into separate actions (on demand, can be scheduled). It hooks documentation generation to the push event and separates the publish action (on demand, could be hooked up to the release event).

Remaining tasks: fix snap and version change. (The next PR will make ci.yml obsolete.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,13 +127,13 @@ jobs:
       - name: Download Docker image
         uses: actions/download-artifact@v4
         with:
-          name: terminusdb-server-docker-image
+          name: terminusdb-server-snapshot-amd64
 
       - name: Generate man page
         run: |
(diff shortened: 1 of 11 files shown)


hoijnet:
The self-hosted runner has been there because of some integration test timing issues with the build. Great if we can remove it, and just adding the comment to help remember if we get the issues back. Fingers crossed! We'll add it back if necessary! Can we remove the self-hosted runner, or is it expected to be used still?

hoijnet (approved):

dfrnt-SigmundFuchs:
>  We'll add it back if necessary! Can we remove the self-hosted runner, or is it expected to be used still?

I’ll continue with this - a few actions are still missing, and the plan is to remove the self-hosted runners.

PR #2207, opened 2025-05-06
Use the debian:bookworm-slim image as the runtime base

dfrnt-SigmundFuchs:
The aim of this PR is to reduce the image size by using the bookworm-slim image as a runtime base and including only handpicked artifacts.


hoijnet (changes requested):
To prevent wild behaviours and ensure that we have predictable builds, specific swipl versions (at least minor versions) should be used if possible so that proper testing can be performed before builds are created so that the jump to 9.4 does not happen before it has been properly tested. The patch releases make sense to follow organically.

Looks great overall!

hoijnet (approved):

hoijnet:
You can decide if the versions should be adhered to. Preventing major jumps is a suggestion.

PR #2211, opened 2025-05-07
Add Bump App Version Action

dfrnt-SigmundFuchs:
This PR extracts a GitHub Actions workflow to automate the process of bumping the application's version number. The workflow allows the version to be updated manually via `workflow_dispatch`, with the new version being written to a `/VERSION` file. It also ensures that only valid version formats are accepted (e.g., `1.0.0`, `1.0.0-dev`). The version update is automatically committed and pushed to the `main` branch.

This action simplifies version management by ensuring the `/VERSION` file is kept in sync with the latest version, streamlining the process for future releases. The workflow is only triggered for the `main` branch to avoid unintended version bumps in feature or development branches. Additionally, it could be triggered automatically after a release to keep the version number up to date with each release cycle.


hoijnet (changes requested):
These changes look good and are approved.

`docs/release-steps.md` is autogenerated and the steps in it are required to follow to make a release currently/previously. I assume that the original release process is taken into account. I could not see a reference to it or handling the process of generating that file or updating the generation of that file with the new information.

Please align and update in/with/on that file generation if necessary so that it is taken into account and the release process is pre-aligned.

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Update
+        run: |
+          echo "Set version to ${{ steps.new-version.outputs.version }}"        
+          echo "${{ steps.new-version.outputs.version }}" > VERSION
+          pip install cogapp
+          cog -v
+          cog -r \
+            -D LAST_RELEASED_VERSION=${{ steps.version.outputs.version }} \
+            -D CURRENT_REPO_VERSION=${{ steps.new-version.outputs.version }} \
+            docs/release-steps.md \
```
filip26, line 81: @hoijnet what about this?

dfrnt-SigmundFuchs, line 81: @hoijnet  what about this?

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,93 @@
...
+            src/config/terminus_config.pl \
+            distribution/snap/snapcraft.yaml
+
+      - name: Commit
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run: |
+          git config user.name "Bump Version Bot"
+          git config user.email "actions@github.com"
+          git add VERSION docs/release-steps.md src/config/terminus_config.pl distribution/snap/snapcraft.yaml
```
dfrnt-SigmundFuchs, line 90: @hoijnet and this?

hoijnet (approved):

alexander-mart:
When will the release be? :eyes:

dfrnt-SigmundFuchs:
@alexander-mart I can't promise a date, but I hope version 11.1.15 will be released within the next two weeks.

Issue #2168, opened 2025-04-09
Snap build is failing due to tests failing - unbundle for existing db


hoijnet:
**Describe the bug**
Unbundle for existing db fails, it seems the tests are incorrect, where a database was not removed before being unbundled again.

Example:
https://github.com/terminusdb/terminusdb/actions/runs/14364837188/job/40276215744

**To Reproduce**
1. Run the snap tests
2. Get error, see link above

dfrnt-SigmundFuchs:
If this suddenly starts failing, the first candidates to check are URLs, e.g., `http://terminusdb.com/schema/api#`.

dfrnt-SigmundFuchs:
Actually, I was wrong — it's been failing for a long time.

```
channels:
  latest/stable:    v11.1.9  2023-11-15 (32) 38MB -
  latest/candidate: ↑                             
  latest/beta:      ↑                             
  latest/edge:      ↑  
```

dfrnt-SigmundFuchs:
@GavinMendelGleason @luke-feeney @matko 

Please help us resolve this issue. The snap was last updated years ago, and I suspect the integration tests for the snap have been failing since then. Thank you.


**Failing tests:**
```
  cli-bundle
>>> ./terminusdb.sh store init --force
{ stdout: 'Successfully initialised database!!!\n', stderr: '' }
>>> ./terminusdb.sh db create admin/db-fbcac5863a81494e48e5dfac7c776c44
{
  stdout: 'Database created: admin/db-fbcac5863a81494e48e5dfac7c776c44\n',
  stderr: ''
}
>>> ./terminusdb.sh bundle admin/db-fbcac5863a81494e48e5dfac7c776c44
    1) passes bundle, unbundle empty database
>>> ./terminusdb.sh db create admin/db-fbcac5863a81494e48e5dfac7c776c44
    2) passes bundle, unbundle database with doc
    bundle, unbundle different empty database
>>> ./terminusdb.sh db create admin/db-fbcac5863a81494e48e5dfac7c776c44
      3) "before all" hook for "passes"
>>> ./terminusdb.sh db delete admin/db-fbcac5863a81494e48e5dfac7c776c44
{
  stdout: 'Database deleted: admin/db-fbcac5863a81494e48e5dfac7c776c44\n',
  stderr: ''
}

  cli-clone-push-pull
    4) catches push has no head

  remote-auth
    5) passes add, get, update, delete, get
    6) deletes with post payload
```

[...]

PR #2212, opened 2025-05-07
Add Publish Actions and Refactor Snap Build

dfrnt-SigmundFuchs:
Introduces separate actions to release the snap, images, and to build the snap. This allows for adding a general `release-all` action, other composite workflows built on top of it, etc.

Let's do the next release in steps together to debug it.


dfrnt-SigmundFuchs:
The latest snap-publish action supports releasing X.X.X-dev versions to the edge channel. Given that it was last updated a year ago, let's publish the snap once issue #2168 is resolved.

hoijnet (approved):
The integration tests have been failing there for a long time so approving as the rest is ok. There will be a need to understand the details of why they are failing. But that will be for a different issue and PR.

PR #2222, opened 2025-07-17
Release notes for 11.1.16

hoijnet:
Following the release process for 11.1.16.


hoijnet:
The bump workflow did not make sense. The "head" option is not supported, and seemed to have been thought to be for "base", which is not required as it is the current branch so should not be needed.

PR #2253, opened 2025-10-27
Fix docker local cleanup

hoijnet:
There are problems building the images when including both root and noroot images, this should now be fixed


dfrnt-AstroFabi (approved):
lgtm

PR #2296, opened 2025-11-25
feat(graphql): add _count query field for counting filtered results

Tails:
Add a new _count field to the GraphQL Query root that allows counting documents matching a filter without returning all the data.

Usage:
  { _count(ModelName: { field: { eq: value } }) }
  { _count(ModelName: {}) }  # count all

The implementation reuses existing filter infrastructure - each model name becomes an argument that accepts its existing filter type.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -21,6 +21,13 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Extract version from source
+        id: version
+        run: |
+          VERSION=$(grep "terminusdb_version(" src/config/terminus_config.pl | grep -oP "'\K[^']+")
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "Extracted version: $VERSION"
+
       - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


src/rust/terminusdb-community/src/graphql/query.rs
```diff
@@ -1338,6 +1338,33 @@ pub fn run_filter_query<'a>(
...
+    let count = lookup_by_filter(
+        context,
+        g,
+        class_name,
+        all_frames,
+        Some(filter),
+        None,  // no zero_iter
+        true,  // include_children (matches default query behavior)
+    )
+    .unique()
```
hoijnet: I'm unsure about the unique here, and should test with real world scenarios before committing so that I know the semantics are accurate. At first glance it looks like unique values, are there valid cases when the count should include more than one value that is the same?

Tails: Good point. I guess actually this would be a pitfall. The count wraps any filter query so if the query only selects a single field that is duplicated across the result items this will already be wrong. It should probably directly rely on the filter itself and not try be smart about it.

PR #2313, opened 2025-12-10
Re-enable the in-memory server

hoijnet:
As reported here, the in-memory server had not been checked for a long time. The server no longer started with that option. It is now in place again. 

https://discord.com/channels/689805612053168129/689886282830315563/1448012161455493131

There is some [preliminary work to a bugfix](https://github.com/ParapluOU/terminusdb/commits/fix/memory-mode-system-db-check/) that helped understand the issue in detail by @Tails


hoijnet:
Made it a local branch instead. Closed this PR:

Issue #2301, opened 2025-11-27
Mitigate Rust Layer Cache Weak-Ref Accumulation Across Branch-Heavy Workloads


hoijnet:
**Is your feature request related to a problem? Please describe.**
Running workloads that churn through many short-lived branches gradually slows every repository on the server. Our Rust-layer cache (a global `LockingHashMapLayerCache` keyed by layer IDs) stores `Weak<InternalLayer>` entries but never proactively removes them. Each branch create/delete cycle leaves behind more dead weak references, so lookups spend increasing time scanning empty slots and attempting `Weak::upgrade()` calls. Once this hash map grows into the tens of thousands of entries, queries and document operations can end up 4–10× slower until the server is restarted, even for databases that never touched those branches.

**Describe the solution you'd like**
Add automatic cleanup to the cache in terminus-store so stale weak references are pruned without depending on restarts. Two viable approaches have already been outlined: (1) augment the existing `LockingHashMapLayerCache` with periodic `retain`-style sweeps triggered every N operations, or (2) replace it with the "Smart Layer Cache" design that uses an LRU/time-based eviction policy and bounded memory. Either path should keep cache lookups close to O(1) regardless of branch churn and avoid cross-database performance bleed-through.

**Describe alternatives you've considered**
1. **Operational workaround** – restart the server occasionally. This restores performance but is brittle and disruptive.
2. **Process changes** – reduce branch churn by batching deletions or keeping branches longer. This helps some teams but doesn’t solve the underlying cache behavior and is hard to enforce across tenants.

**Additional context**
- The cache is instantiated once per server (`CachedLayerStore::new(... LockingHashMapLayerCache::new())`), so any busy repository impacts every other database on the instance.
- Heavy workloads show extra latency per run during very heavy loads
- DFRNT now has reproduction scripts using internal data to validate a fix once the cache is updated.

PR #2342, opened 2025-12-23
Fix eviction of memory for stale layers for #2301

hoijnet:
There were several issues related to inserting many documents in batches, that each contributed to increasing insertion latency in large consecutive runs, consuming memory and having to perform layer traversals.

Performance degradation was observed when performing repeated cycles of INSERT/DELETE/OPTIMIZE operations. Over 20 cycles with 70 documents each:
- Worst case INSERT times was previously degraded from ~45ms to ~280ms (**6.3x slowdown**)
- The degradation reset to ~45ms after server restart, which confirmed this was an in-memory cache issue

This PR clears relevant caches automatically on roll-up, and invalidates previously loaded layer chain, replacing it with the rolled up chain in memory (in terminusdb-store) to ensure continued good performance. Also requires the bumped terminusdb-store version.

### Root Cause Discovery

The issue was traced to the Rust layer cache in `terminusdb-store`. When `exponential_rollup_strategy` performs partial rollups:

1. Intermediate layers get rolled up and their caches are invalidated
2. **BUT** the topmost layer's full parent chain was not invalidated
3. Old ancestor layers remained cached with Arc references
4. Each optimization cycle added more stale cache entries
5. Layer loading had to traverse increasingly deep cached parent chains

The key insight: `register_rollup` in `CachedLayerStore` only invalidated the single rolled-up layer, not its ancestors. This left "orphaned" parent layers in cache that were still reachable via Arc references (making the calls from swipl traverse the longer and longer non-rolled up chain which was not cleaned up).

## Additional issues were also detected and fixed

However, to find the above issue, a number of other contributing factors were also cleaned up and fixed.

Practically, all of the following issues were identified and remediated:
* Dead entries in the layer LRU table lookup cache were not cleared out (now cleared when beyond threshold)
  * There was already a comment about it, fixed it at the same time
* Caches on the Swipl side were not properly cleaned up (now using retractall)
* Commit graph delta roll-up was not in properly setup (causing commit graph buildup)
* TerminusDB Store did roll up, but did not replace in-memory layers with roll-up, see above main fix

### Layer cache architecture, via terminusdb-store

```
┌─────────────────────────────────────────────────────────┐
│ LockingHashMapLayerCache                                │
│ HashMap<[u32;5], Weak<InternalLayer>>                   │
│                                                         │
│ Layer A ←── Layer B ←── Layer C ←── Layer D (topmost)   │
│   ↑            ↑            ↑            ↑              │
│  Weak        Weak         Weak         Weak             │
│                                                         │
│ Problem: When B is rolled up, only B was invalidated.   │
│ A remained cached, still referenced by B's Arc.         │
│ Fix: Invalidate A, B, C, D - the entire chain.          │
└─────────────────────────────────────────────────────────┘
```

## Before optimizations (build a clearer test)

===== IN-MEMORY CACHE ACCUMULATION TEST =====
  This test measures INSERT timing across cycles.
  Run this test, then restart server (NO --clean), then run again.
  If degradation resets after restart, the issue is in-memory caching.

  Creating new database: restart-test-db
  Cycle 1: INSERT=46ms, DELETE=2301ms, OPTIMIZE=37ms
  Cycle 5: INSERT=86ms, DELETE=3100ms, OPTIMIZE=34ms
  Cycle 8: INSERT=193ms, DELETE=4262ms, OPTIMIZE=32ms
  Cycle 10: INSERT=300ms, DELETE=5228ms, OPTIMIZE=65ms

  ===== RESULTS =====
  First INSERT: 46ms
  Last INSERT: 300ms
  Degradation: 6.52x

## After (stable, but measurements fluctuate a little).

```
===== IN-MEMORY CACHE ACCUMULATION TEST =====
  This test measures INSERT timing across cycles.
  Run this test, then restart server (NO --clean), then run again.
  If degradation resets after restart, the issue is in-memory caching.

  Creating new database: restart-test-db
  Cycle 1: INSERT=43ms, DELETE=3047ms, OPTIMIZE=59ms
  Cycle 5: INSERT=47ms, DELETE=2826ms, OPTIMIZE=50ms
  Cycle 6: INSERT=44ms, DELETE=3292ms, OPTIMIZE=63ms
  Cycle 8: INSERT=45ms, DELETE=3126ms, OPTIMIZE=49ms
  Cycle 10: INSERT=50ms, DELETE=2782ms, OPTIMIZE=78ms

  ===== RESULTS =====
  First INSERT: 43ms
  Last INSERT: 50ms
  Degradation: 1.16x (note: not really, these are fluctuations more)
  ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,8 @@ lint-mocha-fix:
 rust:
 	@$(MAKE) -f distribution/Makefile.rust
 
-# Run the unit tests in swipl.
+# Run unit tests in swipl; all, or just one suite.
+# make test OR make test SUITE='[json,terminus_store,tables]'
 .PHONY: test
 test:
 	@$(MAKE) -f distribution/Makefile.prolog $@
@@ -133,4 +134,4 @@ $(ROFF_FILE): $(RONN_FILE)
 	ronn --roff $<
 
 .PHONY: pr
-pr: clean dev restart lint lint-mocha test test-int
+pr: lint lint-mocha clean dev restart test test-int
--- a/distribution/Makefile.prolog
+++ b/distribution/Makefile.prolog
@@ -48,11 +48,15 @@ i: $(RUST_TARGET)
 
 .PHONY: test
 test: $(RUST_TARGET)
+ifdef SUITE
 	$(SWIPL) \
-	  --on-error=halt \
-	  --on-warning=halt \
-	  -g 'run_tests, halt' \
+	  -g '(run_tests($(SUITE)) -> halt(0) ; halt(1))' \
 	  -f src/interactive.pl
+else
+	$(SWIPL) \
+	  -g '(run_tests -> halt(0) ; halt(1))' \
+	  -f src/interactive.pl
+endif
 
 .PHONY: download-lint
 download-lint: $(SWIPL_LINT_PATH)
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -4,8 +4,8 @@ Thanks for taking the time to contribute to TerminusDB!
 
 ## Testing
 
-Before submitting a change, please run `make ; ./terminusdb test` to make sure
-that all tests pass.  Failure should result in a big fail message, and
+Before submitting a PR, please run `make pr` to make sure
+linting and all tests pass.  Failure should result in a big fail message, and
 success with a final `true`. API tests will require that the admin
 password is `root` or that the environment variable
 `TERMINUS_ADMIN_PASSWD` is set prior to invocation of `terminusdb`.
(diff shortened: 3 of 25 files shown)


hoijnet:
Pre-release docker image for this available as `terminusdb/terminusdb-server:2301-accumulation-cache-lru-eviction` 

See: [2301-accumulation-cache-lru-eviction](https://hub.docker.com/layers/terminusdb/terminusdb-server/2301-accumulation-cache-lru-eviction/images/sha256-31c0c57c4d5a4a34d4dd055d7896de1b0d244adcfb4b39bb400d5024abbc12ee)

dfrnt-HansKochstein (approved):

PR #2358, opened 2026-01-17
Fix incorrect term framing in Rust FFI in specific long-running integration test

hoijnet:
A garbage collection off-by-one error ("relocation cells" error) was occurring during heavy document insert/update workloads, particularly when using WOQL patterns involving `opt().dot()` with `update_document`. The crash manifested during optimizer thread creation and appeared to be related to term reference accumulation in the Rust FFI layer. Specifically, it seems to have been connected to newly optimized JSON list processing that did not handle terms framing correctly. 

Investigation revealed that several Rust functions were not properly scoping their Prolog term references. In swipl-rs, term references created with `context.new_term_ref()` persist until the end of the FFI call, while those created within a `Frame` (via `frame.new_term_ref()`) are released when the frame is closed. Without proper frame scoping, term references accumulate over repeated calls, increasing memory pressure and once GC runs, if FFI calls don't handle context properly, GC safety features trigger a crash.

What was happening was that terms were built in frames that were closed before the terms were unified correctly. The context should be used for all terms that are returned to swipl so that they are accounted for via swipl-rs.

### Auto-optimizer plugin

Garbage collection is introduced into the auto-optimizer plugin to keep the memory usage and thus also reference handling lean for consistent performance. Every 100th commit on average will trigger a garbage collection run, and every 10th commit on average will trigger an optimization of the layers, building rollups based on the roll-up patterns.

Allow transaction retries that happens when the auto-optimizer runs, and only warn about the retries happening. Some tests written had introduced transaction retry checking. We allow one retry and fail if more than one now. Also they are echoed out as warnings for information.

### terminusdb-community/src/json_preserve.rs

Added explicit scoping and fast paths for empty collections in JSON-to-Prolog conversion:

- Empty arrays and objects now handlt term reference creation correctly
- Populated arrays and objects handle term reference creation correctly
- Similar treatment for `json_read_all_from_stream` result processing

### core/transaction/database.pl

Removed `abolish_all_tables` from `api_optimize.pl` and `api_rollup.pl` as `post_transaction_tabling` already handles table abolishment. `abolish_all_tables` causes issues with GC as tables are abolished during handling by the GC. This ensures table clearing runs privately in the thread by the transaction (private abolishment that was already implemented), avoiding potential race conditions with concurrent optimizer threads that could trigger GC at unsafe moments.

### core/api/api_optimize.pl and core/api/api_rollup.pl

Removed direct `abolish_all_tables` calls, now handled centrally in `post_transaction_tabling` in each thread.

## Technical Details

The key insight is understanding the swipl-rs frame lifecycle:

1. When `context.open_frame()` is called, terms created with `frame.new_term_ref()` are scoped to that frame
2. After `object_term.unify_arg(1, &f_term)`, the value is copied into `object_term`'s structure
3. Calling `frame.close()` releases the term reference but the unified data persists
4. Without explicit frame closure, term references accumulate until the end of the FFI call
5. The term that is returned in the FFI call must be created by the context, not the frame.

This pattern is especially important for functions called repeatedly in tight loops, such as during batch document processing. A critical insight is that results must be within the context, and not built within a frame.

## Testing

- Build succeeds with `make dev`
- Existing unit and integration tests pass (meke pr)
- Stress testing with high-volume document operations performed to verify the fix

### DIFF
--- a/docker/plugins/auto-optimize.pl
+++ b/docker/plugins/auto-optimize.pl
@@ -13,6 +13,12 @@
 % Copyright (c) TerminusDB, Licensed under the Apache License, Version 2.0
 % Source: https://github.com/terminusdb-labs/terminusdb-plugins/tree/main
 
+% Disable background Garbage Collection thread AND enable automatic GC on plugin load.
+% GC will ONLY run via explicit gc_safe/0 calls which coordinate with Rust FFI.
+:- set_prolog_flag(gc_thread, false).
+:- set_prolog_flag(gc, false).
+:- format(user_error, '% auto-optimize: swipl GC fully disabled, coordinated GC only~n', []).
+
 optimize_chance(system_descriptor{}, C) =>
     C = 0.1.
 optimize_chance(D, C), database_descriptor{} :< D =>
@@ -28,8 +34,8 @@ should_optimize(Descriptor) :-
     random(X),
     X < Chance.
 
-% GC runs on average every 100 commits (1% chance)
-gc_chance(0.01).
+% GC runs on average every 10 commits (10% chance)
+gc_chance(0.1).
 
 should_garbage_collect :-
     gc_chance(Chance),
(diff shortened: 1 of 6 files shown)


hoijnet:
This has been obsoleted by later work in swipl-rs and the support for swipl-10 in #2366. Closed again. The root cause  seems to have been improper frame handling for integers in swipl-rs that is now being addressed.

PR #2362, opened 2026-01-20
tests: respect TERMINUSDB_EXEC_PATH override

daniel-fahey:
This PR makes specific test files prefer `TERMINUSDB_EXEC_PATH` when spawning the TerminusDB CLI, while keeping the existing in-tree `./terminusdb` default when the variable is unset.

Updated:
- `tests/test/cli-version.js` (env passed to `terminusdb.sh --version`)
- `tests/test/memory-mode.js` (binary path used for `spawn(...)` in the memory-mode tests)

This helps downstream packagers and CI environments run the JS integration tests against an installed binary, wrapper script, or non-standard build output location without patching the repository.

Context: originally identified while packaging TerminusDB in Nixpkgs: https://github.com/NixOS/nixpkgs/pull/303209

### DIFF
--- a/tests/terminusdb.sh
+++ b/tests/terminusdb.sh
@@ -49,5 +49,5 @@ elif [[ $use_exec -eq 0 ]]; then
   if [ -t 1 ]; then
     set -x
   fi
-  TERMINUSDB_LOG_LEVEL="ERROR" "$TERMINUSDB_EXEC_PATH" "$@"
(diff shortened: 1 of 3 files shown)


hoijnet (approved):
Updated to use the standard framework. Hope ok. Merging these as they should be fixed anyways. Please retest to ensure all works as expected!

PR #2435, opened 2026-04-27
Duplicate: Fix/swap value subdocument

hoijnet:
Based on @picbeats PR that fixes https://github.com/terminusdb/terminusdb/issues/2405
There seems to be an issue on github currently, can't see the relevant PR //Philippe

## Problem

`values_equal/2` in `src/core/document/patch.pl` uses `atom_string/2` to coerce both arguments before comparing them. This works for atoms and strings, but **crashes with `type_error`** when either argument is a SWI-Prolog dict (`json{...}`), because `atom_string/2` does not accept dicts.

**Failing scenario:** A `SwapValue` patch operation where `@before` is `null` (an atom) and `@after` is a subdocument object:

```json
{
  "@op": "SwapValue",
  "@before": null,
  "@after": { "@id": "Entity/e1/quantity/Quantity/1", "@type": "Quantity", "value": 42 }
}
```

This produces an **HTTP 500** on the `/api/diff` and `/api/apply` endpoints whenever a document field changes from `null` to a subdocument object (or vice-versa), e.g. `NDArray` or `Quantity` fields.

## Fix

Add a new `values_equal/2` clause (before the existing `atom_string/2` clause) that handles dicts, lists, and numbers via structural unification with a cut to prevent fallthrough:

```prolog
values_equal(V1, V2) :-
    (   is_dict(V1) ; is_dict(V2)
    ;   is_list(V1) ; is_list(V2)
    ;   number(V1)  ; number(V2)
    ),
    !,
    V1 = V2.
```

The existing atom/string coercion clause is unchanged and handles all remaining cases.

## Tests

Added 7 tests to the `simple_patch` test suite in `patch.pl`:

| Test | Description |
|------|-------------|
| `swap_null_to_subdocument` | SwapValue from `null` to a subdocument dict (the crash case) |
| `swap_subdocument_to_null` | SwapValue from a subdocument dict back to `null` |
| `swap_between_subdocuments` | SwapValue between two different subdocument dicts |
| `swap_string_value_regression` | Normal string SwapValue still works (regression check) |
| `values_equal_dict_identity` | Equal dicts unify correctly |
| `values_equal_dict_not_equal` | Different dicts correctly fail |
| `values_equal_null_vs_dict` | `null` vs dict fails (does not throw) |

## Related

- **Race condition fix (separate PR):** https://github.com/terminusdb/terminusdb/pull/2408 — fixes `create_db` race condition that can cause `context_not_found` errors; tracked in https://github.com/terminusdb/terminusdb/issues/2407
- **Fork PR:** https://github.com/Qruise-ai/terminusdb/pull/1 — Qruise-ai internal fork PR containing this fix


hoijnet:
@picbeats sorry for taking some time to get this fixed. Testing the build with the latest main here to detect any regressions, and if the above are building correctly, we can merge your PR. For some reason I can't see it currently on TerminusDB. Please update your PR with the latest main and I can merge that one instead.

hoijnet:
Closed in favor of updated PR from @picbeats: https://github.com/terminusdb/terminusdb/pull/2438

PR #2437, opened 2026-04-29
@shared class-level annotation for cascade-deleted shared documents

### LINKED ISSUE #2111: Reference-counted types to enable flexible links between objects
**Is your feature request related to a problem? Please describe.**
Today one can either create a document type to constrain RDF triples into a shared lifecycle, or create a subdocument to add a "layer" of information to a type (JSON-LD). 

An elegant solution for bi-directional links would be to have a reference-counted linked relationship between two objects. Imagine the concept of friendship, being between two persons, seeing it as directional gives a skewed view of reality, a bi-directional link as a reference-counted third object that connects the individiual would be quite handy.

Another similar instance is a configuration of items into an offering. It could of course be modelled as a type with a set of links, either as individual properties, or as a set of link with a Cardinality. However, if any of the parts of the offering is removed, the offering could be relevant to automatically delete as well (as a design choice).

**Describe the solution you'd like**
Having reference counted type properties (if one of the property linked is removed, the entire object should be removed too) would enable a significant opportunity to avoid having to write custom logic and instead have more operations possible to perform over the REST and GraphQL interface.

**Describe alternatives you've considered**
WOQL logic for it, but would be hard for the use case.

**Additional context**
N/A

hoijnet:
This adds `@shared` as a class-level annotation in TerminusDB. A `@shared` class creates standalone documents (with their own IRI) that are automatically cascade-deleted when the last reference to them is removed. This is a middle ground between regular documents (no cascade) and subdocuments (embedded, single-owner).

Importantly, `@shared` documents (conceptually similar to `@subdocument`) can be created as standalone documents and later linked from other documents. But then, when the last document that links to it is deleted, the `@shared` document is deleted too. The design principle behind it, is to allow free-form documents to be created and later linked for multi-transaction flows where loading happens sequentially, such as in a branch that will later be merged as one commit.

Intentionally, there is no `@child` that can be specified on a property as the semantics of that would be unclear and the performance implications may be substantial, at least in this initial implementation. What should the lifecycle be, and the expectations on which documents would be in scope for scanning is non-trivial. The `@shared` annotation is closer to the original `@subdocument` and relatively easy to reason about.

### Use case

This is for documents that are shared between multiple parts of one or more documents, which we have found is a relatively common design pattern across customers and our own work. The workarounds have not always been pretty, and the same needs come up time and again. 

A `@shared` document can be referenced by multiple parent documents and persists as long as at least one reference exists. When the last referencing parent is deleted, the shared document is automatically garbage-collected in the same transaction. The deletion is atomic — either all cascade deletes succeed or the entire transaction rolls back.

### Schema (`src/core/document/schema.pl`)
- Recognises `@shared: []` class annotation
- `is_shared_class/2` predicate exposes the annotation to Prolog
- `schema_is_shared/2` walks the schema layer for any shared class
- `schema_is_shared_tabled/2` — tabled predicate for O(1) repeated lookups within a transaction

### Document API (`src/core/api/api_document.pl`, `src/core/document/json.pl`)
- `get_shared_ids_from_schema/2` — collects all `@shared` class IRIs at DocumentContext construction time
- Shared documents are treated as regular (standalone) documents for all read/write operations; cascade delete is the only distinction

### Cascade delete (`src/core/plugins.pl`)
- `pre_commit_hook/2` multifile hook — fires after validation, before durable commit
- Two lightweight O(1) guards before invoking cascade:
  - `has_shared_schema_types/1` — checks tabled schema result
  - `has_any_instance_deletions/1` — `xrdf_deleted/4` stops at first match
- `perform_shared_cascade/2` calls `'$doc':cascade_shared/3` (Rust) in a loop until no new orphans are found
- Falls back to pure Prolog path if the Rust extension is unavailable

### Rust fast path (`src/rust/terminusdb-community/src/doc/delete.rs`)
- `cascade_shared_deletes/3` Prolog foreign predicate (`'$doc':cascade_shared/3`)
- Walks all deleted subjects; for each, finds all objects that are `@shared` instances
- Liveness check: `is_shared_target_live_memo` — uses a `HashMap<u64, bool>` memo table shared across all candidates in one pass to avoid redundant backward-reference scans
- Orphan batch filter: builds `HashSet<u64>` of orphan node IDs before unlinking, so intra-orphan backward references are skipped during the unlink phase
- Returns count of deleted orphans to the Prolog loop

### ACID atomicity

The `nb_commit` calls inside `transaction_objects_to_validation_objects` flush the Prolog layer builder to an **in-memory** layer object — no disk write occurs. The durable write happens only in `commit_validation_objects` at the end of a successful `with_transaction`. If cascade fails at any point the exception propagates, `with_transaction` does not call `commit_validation_objects`, and all in-memory layers are garbage-collected. Full ACID atomicity is preserved.

---

## Test coverage

**File:** `tests/test/shared-child-annotation.js` — 39 tests across 8 `describe` blocks

| Block | Scope |
|---|---|
| basic `@shared` instance creation | schema validation, IRI format |
| reference counting (single referrer) | cascade on last-ref delete |
| multi-referrer cascade | partial delete leaves shared doc alive |
| collection element cascade | Set fields, per-element cascade |
| standalone creation and update | explicit create/update without parent |
| error cases | schema violations, invalid annotations |
| data integrity — preservation | surviving doc intact; bystander doc untouched; collection intact after cascade |
| data integrity — unrelated data | non-`@shared` documents unaffected by cascade in same transaction |

Key preservation tests added:
- **Surviving referencing document is intact after partial cascade** — verifies all fields (`@id`, `@type`, `title`, `notes` array with correct IRIs) on the non-deleted referrer
- **Unrelated bystander document fully intact** — verifies a document with no `@shared` references is untouched after a cascade in the same DB
- **Parent collection intact after element cascade** — verifies the collection document itself and its `items` Set after cascade removes some elements

---

## Known limitations

- Cascade loop is bounded to **≤5,000 documents per request** for reliable performance. Beyond 10,000 documents in a single cascade the server enters a super-linear regime and blocks all other requests for the duration (better performance than `@subdocument`, but still a limitation).
- The cascade runs on the Prolog dispatch thread (single-threaded); large cascades are not interruptible.
- `DocumentContext::new()` scans the schema for `@shared` classes on every document operation. In databases that use `@shared` extensively, unrelated deletes pay a small O(M) Rust scan where M is the number of deleted triples. This is a pre-existing cost of the context initialisation, now with real work behind it.

## No breaking changes

The `@shared` annotation is opt-in. Databases without `@shared` classes are unaffected: the tabled schema check returns false immediately and the cascade hook exits without scanning.

## Related

There may be optimizations possible for both `@shared` and `@subdocument` deletion to explore as they are similar in nature and have approx O(N²) complexity which is suboptimal (since previously).

Documentation is being updated with information about `@shared` too.


hoijnet:
There is more work on this one, a couple of more tests are underway:
* Inheritance of `@shared` to block `@subdocument`, and actually be inherited as well
* A few additional potential corner cases (optional document field removes link, duplicate set references)

hoijnet:
Added tests:
  - Update-triggered cascade (field re-point and Optional cleared)
  - Multi-hop cascade through @shared chains
  - Duplicate Set reference deduplication
  - Circular @shared island collection
  - Subclass inheritance of @shared cascade
  - Schema validation rejecting @shared/@subdocument mixing in inheritance chains
  - Cascade through @subdocument containing @shared ref
  - Liveness: subdocument reference keeps @shared target alive
  - No false cascade through regular document boundaries
  - Confirms cascade is triple-removal driven, independent of @unfold/@unfoldable

hoijnet:
@Brian-Stovall-contractor For the built docker container this exists to test. Should be ready to merge:
`docker.io/terminusdb/terminusdb-server:feature-shared-child-annotation`

Brian-Stovall-contractor (approved):

PR #2450, opened 2026-06-22
 Fix silent document loss for non-ASCII characters in document API

Brian-Stovall-DFRNT:
## Description
  
  This was a subtle combo of two errors: the default Latin-1 encoding of prolog, plus a blanket catch that silently returned a [] with 200 status instead of a descriptive error. This didn't often surface because most interactions go via WOQL, which doesn't have this error in its path. Also there was an extant test that was getting 'covered' by the test harness, so not failing as it should have.
  
  When a client sends Content-Type: application/json; charset=utf-8,
  SWI-Prolog's http_read_data decodes the HTTP body into Unicode code
  points. open_string/2 then creates an iso_latin_1 stream, so the Rust
  JSON parser receives Latin-1 bytes rather than valid UTF-8 — a ü
  (U+00FC) arrives as the single byte 0xFC instead of 0xC3 0xBC. serde_json
  throws a parse error, which a blanket catch(_Error, fail) converts to
  EOF, silently returning [] with no error.

  Fixed the encoding by re-encoding the buffered string to raw UTF-8 bytes
  via a memory file before handing it to the Rust parser. Also propagate
  parse errors from json_read_dict_stream rather than swallowing them, so
  malformed input returns a proper HTTP error instead of an empty response.

  Add regression tests in tests/test/document-utf8.js covering both the
  charset=utf-8 trigger path and the no-charset control case.

## In Depth:

See my originating ticket: https://github.com/dfrnt-com/twinfoxdb-enterprise/issues/46

### DIFF
--- a/src/core/util/json_preserve.pl
+++ b/src/core/util/json_preserve.pl
@@ -65,7 +65,7 @@ json_read_dict(Stream, Dict, Options) :-
 json_read_dict_stream(Stream, Dict, Options) :-
     % Use Rust parser for one-at-a-time streaming
     % All streams are now string streams (buffered from HTTP)
-    (   catch('$json_preserve':json_read_one_from_stream(Stream, Value), Error, throw(Error))
+    (   '$json_preserve':json_read_one_from_stream(Stream, Value)
     ->  Dict = Value
     ;   % Rust parser returned nothing - must be EOF
         handle_eof(Options, Dict)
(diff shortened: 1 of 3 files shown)


hoijnet:
Thanks for the PR. Adjusted it to a simpler and more efficient path. Should be good from here.

hoijnet (approved):
Bug squashed! Thanks Brian!!

hoijnet:
Thanks for the contribution!! The latest 12.0.7 release is now out since today!

PR #2457, opened 2026-06-30
Parallel and sequential write improvements, with revised elaboration, commit queue, and contract-based verification

hoijnet:
Profiling the document write path revealed that JSON elaboration — the step where an incoming JSON document is type-checked against the schema, properties are expanded, and IDs are assigned — dominates total insert time. For a batch of 10 000 simple documents it accounts for roughly 84 % of wall-clock time. Everything after that — triple materialization, the layer builder, the commit graph update — is comparatively cheap.

The existing write path was single-threaded: one request thread elaborates every document in sequence, then opens a transaction and commits. For workloads that send large batches this means all but one CPU core are idle during the most expensive phase. Worse, the serialized approach means that concurrent writers contend heavily on the single commit point, and with the `with_transaction` retry mechanism for the document API, a writer that lost the race would drive contention with heavy re-elaboration costs; often failing multiple times.

This PR is the first concrete step toward fixing both problems, it uncovers significant improvements in parallel write performance and lays a path for continued write performance improvements.

### 1. Parallel elaboration queue

A new Prolog module (`src/core/document/parallel_elaboration.pl`) implements a work-stealing queue that farms out elaboration work to a pool of idle Prolog worker threads. The request thread splits the incoming document batch into chunks, enqueues them, and each worker thread elaborates its chunk independently. Because elaboration is almost entirely read-only schema work, the workers all share the same schema snapshot without needing any locks.

The result comes back as a list of pre-elaborated documents. The request thread then opens a short transaction, inserts the pre-elaborated documents, and commits. The transaction window shrinks from ~2 200 ms to sub-second latency for a 10 000-document batch, depending on worker count (in this case 8 cores).

### 2. Rust elaboration fast path

For the simplest eligible documents — flat objects with a `Random` key and primitive `xsd:*` properties, no explicit `@id`, no captures, no nested documents — a new Rust predicate (`$doc:rust_elaborate_simple_documents/3`) handles the elaboration entirely in Rust. The output is identical to the Prolog path; the Rust path simply gets there much faster. This is a minuscule first step of optimizing the elaboration using rust for beneficial cases.

If a document is not eligible (captures, subdocuments, explicit ID, complex key types) the Rust path fails fast and the worker falls back to the standard Prolog elaboration transparently.

### 3. Prefix expansion moved to Rust

Prefix checking and expansion — mapping `@type` values and property names against the schema context — is done via a Rust-backed predicate rather than a Prolog loop over the prefix list. This removes a significant per-document Prolog overhead for schemas with many custom prefixes. This was an important step to prepare for moving more work to the rust side later.

As this is a core feature used across all parts of TerminusDB, this could also improve performance elsewhere.

### 4. Commit queue with per-thread wake semantics

A new commit queue (`src/core/document/commit_queue.pl`) serializes the transaction commit phase across concurrent requests. This is the part of the write path that cannot be parallelized: only one commit can move the branch head at a time.

The key design decision is that workers do not poll or sleep. Instead, each worker thread owns a personal message queue and blocks on it until the commit coordinator sends it a wake message. This keeps idle threads off the CPU and makes the wake latency predictable.

There are two ways a transaction reaches the queue in the initial implementation:

**Retry mode** — the request elaborates, tries to commit, loses the race (the branch head moved while it was elaborating), receives a `fail_transaction` signal, and re-enqueues itself. The retry count is exposed as a response header so callers can observe backpressure. (this retry mode was later abolished entirely in favor of the flow mode).

**Flow mode** — the request hands the elaborated batch directly to the queue without attempting a speculative commit first. The queue's single commit worker processes requests one at a time in arrival order. This avoids wasted elaboration on writes that are very likely to conflict (e.g. a tight loop of small writes on the same branch). Also, the ordering of commits is now upheld.

### 5. Contract-based elaboration verification

The elaboration work is now moved to a pre-commit step to enable small writes to cut the queue and move past larger batches, which keeps the latency low on average. However, this means that the world may move between the start of the elaboration for a large import and the completion of it.

Thus, contract-based elaboration verification was introduced where a commit contract is established tracking all dependencies that needs pre-commit checking. Before an elaborated batch is committed, the system checks whether anything changed in the commit graph since the schema snapshot was read for elaboration. This is expressed as two sets of triples:

- **Must-exist triples**: referenced documents that the elaboration assumed were present.
- **Must-not-exist triples**: IDs that the elaboration assumed were free or to not exist.

A rolling Rust-backed change window tracks every commit that has landed since the earliest in-flight request began. If the change window does not intersect either verification set, the pre-elaborated batch is still valid and can be committed immediately. If it does intersect:

- Schema layer changed → full synchronous retry (the snapshot is stale in a fundamental way).
- Only generated IDs affected → regenerate IDs, rematerialize the affected triples.
- References, captures, or subdocuments affected → re-elaborate those documents against the latest schema.

This keeps the common case (no conflicts) extremely cheap — a Rust set intersection — and reserves the expensive retry work for the cases that actually need it. This means that small independent commits can flow freely and with low latency.

The change-window implementation is in `src/core/document/commit_queue.pl` and the Rust side in `src/rust/terminusdb-community/src/doc/change_window.rs`.

It should be noted that the commit sequence is per branch. Each branch gets its own commit queue and work is scheduled fairly across branches.

### 6. Commit-window guard for the existing document API

The pre-commit hook checks the change window before any transaction commits using the commit contract. For the CLI and single-request paths that do not use the worker pool, commits use the old retry mode.

Each worker runs a loop that tries, in order: steal an elaboration chunk from any in-flight request, then try commit work, then block waiting for a wake signal. The balance between elaboration stealing and commit work is tunable. 

The `TERMINUSDB_WORKER_ELABORATION_PREFERENCE` environment variable (a float from 0.0 to 1.0, default 0.5) controls the distribution and the preference for a worker to try elaboration stealing over commit work on each iteration of its loop. A value of 1.0 means always try elaboration first; 0.0 means always try commit work first. Since each elaboration work is at least 1000x faster, tuning towards 0.9 or more may be relevant in some cases. This requires some future exploration.

There is also a built-in anti-starvation rule: if a commit has been waiting in the queue for more than 2 seconds, the preference is overridden and the worker commits first regardless of the configured bias.

## How the previous strategy differed

Previously, a document insert request did everything sequentially and alone:

1. Open the branch descriptor (snapshot the schema and instance layers).
2. For each document: elaborate in Prolog, materialize triples into a layer builder.
3. Commit: build the layer, update the commit graph, move the branch head.

If two requests arrived at the same time and both opened the same snapshot, the first to commit would succeed. The second would fail at `nb_set_head` (the Rust layer store's atomic head-update check), and retry, sometimes many times and ultimately in some cases return an excessive retry error.

In that design, under concurrent write load, throughput does not scale with cores and failure rate climbs with concurrency.

## What this PR does not yet change

Rust elaboration eligibility covers only the simplest document shape today. Extending it to cover `ValueHash`/`Lexical` keys, optionals, sets, and subdocuments is future work based on actual performance benefits.

## Testing

New PLUnit suites:

- `commit_queue` — enqueue, dequeue, process-one-commit, and pool lifecycle.
- `commit_queue_helpers` — backoff, wake semantics, and queue state predicates.
- `parallel_elaboration` — Rust/Prolog output parity for eligible documents.
- `parallel_initial_writes` — end-to-end concurrent insert contracts.
- `json_preserve` — JSON stream tail-closure and error handling.


hoijnet:
Merged into 12.1-rc branch instead.

PR #2461, opened 2026-07-04
Extend commit queue concept and fix race between rebase and auto-optimizer on `_meta` graph

### LINKED ISSUE #2446: Transient 500 on /api/rebase: background _meta auto-optimization races in-flight rebase (v12.0.4 and v12.0.5)
## Summary

`POST /api/rebase/<org>/<db>/local/branch/main` intermittently returns a transient `500 {"api:message":"Unexpected failure in request handler","api:status":"api:failure"}` when the server's background `_meta` layer auto-optimization completes immediately before/during the rebase. The identical request succeeds on retry. I reproduced this on **v12.0.4 and v12.0.5** — so #2400 ("Only run auto-optimize on write and with parent") doesn't appear to cover this window; the rebase flow itself writes, so optimization still runs around it.

## Environment

- Images: `terminusdb/terminusdb-server:v12.0.4` and `:v12.0.5` (official Docker Hub)
- Single-node, default configuration, `TERMINUSDB_ADMIN_PASS` set, no clustering
- Reproduces most readily under CPU pressure (I used `docker update --cpus 0.3`) — which matches how it first surfaced for me: only on shared CI runners, never on a fast local machine

## Reproduction

Loop the following sequence against a CPU-throttled server. My failure rate: ~15–20% of iterations at 0.3 CPU; 0% unthrottled across 150 iterations.

```bash
docker run -d --name tdb -p 6363:6363 -e TERMINUSDB_ADMIN_PASS=root \
  terminusdb/terminusdb-server:v12.0.5
docker update --cpus 0.3 tdb
```

Per iteration (fresh DB name each time, e.g. `race_$(date +%s%N)`):

1. `POST /api/db/admin/<db>` — create database
2. `POST /api/document/admin/<db>/local/branch/main?graph_type=schema&author=system&message=init` — insert a simple document class
3. `PUT /api/document/admin/<db>/local/branch/main?author=t&message=seed&create=true` — upsert 2–3 documents on main (a few separate commits help — they give the auto-optimizer work)
4. `POST /api/branch/admin/<db>/local/branch/feature` with `{"origin": "admin/<db>/local/branch/main"}`
5. `POST /api/document/admin/<db>/local/branch/feature?author=t&message=branch-doc` — write a doc on the branch
6. `POST /api/rebase/admin/<db>/local/branch/main` with `{"rebase_from": "admin/<db>/local/branch/feature", "author": "t"}` ← **intermittent 500 here**
7. `DELETE /api/db/admin/<db>`

## Evidence

Client-side capture:

```
POST /api/rebase/admin/learning_itest_mqbdwa7r/local/branch/main
→ 500 Internal Server Error
→ body: {"api:message":"Unexpected failure in request handler","api:status":"api:failure"}
(immediate retry of the identical request → 200)
```

Server DEBUG log — the optimization completes immediately before each 500, every time:

```
[DEBUG] 2026-06-12T20:33:41.279416+00:00 Optimized admin/learning_itest_mqbdwa7r/_meta
[INFO]  2026-06-12T20:33:41.284581+00:00 POST /api/rebase/admin/learning_itest_mqbdwa7r/local/branch/main (500)
[DEBUG] 2026-06-12T20:33:41.537275+00:00 Request fff57a70-... started - POST /api/rebase/... (retry)
→ retry returns 200
```

I also observed a correlated optimization-side failure in the same sessions, suggesting layer files being swapped/removed under concurrent readers:

```
[ERROR] Optimization of admin/<db>/_meta failed: error(rust_io_error(NotFound,No such file or directory (os error 2)),_1462)
```

Counts from my runs: v12.0.4 — 9 rebase 500s / 50 throttled iterations; v12.0.5 — 5 / 30. Every other operation in the loop (db create/delete, branch create/write/delete, document CRUD) never failed across thousands of executions; the 500 is exclusive to `/api/rebase`.

## Expected behavior

Either the rebase succeeds, or (if a retryable internal conflict with optimization is detected) a structured, identifiable error rather than a generic unhandled-exception 500 — so clients can distinguish "retry me" from "real failure".

## Offer to contribute

I'm happy to:

1. **Contribute this reproduction as a regression test** in your test suite (I already run it in my CI shape and can adapt it to yours), and
2. **Work on the fix with guidance** — my read is that the auto-optimizer's layer-file swap isn't synchronized with the rebase path's layer reads, but you know the storage-engine invariants; point me at the right seam and I'll attempt a patch.

Context: TerminusDB is the VCS backend for my platform's entity version history, so I have a standing interest in the branch/rebase path being robust — thanks for the work on it.

hoijnet:
This PR builds on #2457 that should be merged first.

`POST /api/rebase/<org>/<db>/local/branch/main` intermittently returned HTTP 500 with `Unexpected failure in request handler`. The failure correlated with the auto-optimizer running on the same database immediately before the rebase. The root cause was a race on the `_meta` graph head.

The rebase path in `db_rebase.pl` updates the repository head in the `_meta` graph and then commits the database transaction with `run_transactions/3`. The auto-optimizer, triggered by `post_commit_hook` from an earlier commit, runs in a detached thread and squashes the `_meta` graph head using `nb_force_set_head/3`. Squashing creates a new base layer, so the new layer the rebase is about to commit is no longer a child of the current head. `nb_set_head/2` in the database commit path then fails silently, the rebase predicate fails, and the HTTP handler returns a generic 500.

## Improvements

In addition to fixing the race issue, a queue processing worker issue was fixed, improving some parallel processing scenarios further than the PR baseline this PR is based on. The `bench.py` file in `tests/benchmark/congestion_document_insert` can be used to test the parallel write performance baseline for simple documents. For 80k documents written with 2 writers using 20k chunks; it writes approx. 24000 documents/second now. Sequential writes with a single writer is also quite good with 19k/sec. (most of the improvement work was in the other PR).

```json
{
  "benchmark": "congestion_document_insert",
  "metrics": {
    "writers": 2,
    "chunks_per_writer": 2,
    "docs_per_chunk": 20000,
    "parallel": {
      "total_ms": 3280.2743749998626,
      "documents": 80000,
      "chunks": 4,
      "documents_per_second": 24388.203806885314,
      "ms_per_document": 0.04100342968749828,
      "verified_documents": 80000
    },
    "sequential": {
      "total_ms": 4085.8709160002036,
      "documents": 80000,
      "chunks": 4,
      "documents_per_second": 19579.668972585823,
      "ms_per_document": 0.05107338645000255,
      "verified_documents": 80000
    },
    "congestion_ratio": 0.8028335800219145
  }
}
```

This PR brings more of the write path into the queue-based processing.

## What changed

This fix builds on a separate PR that improves parallel write performance and reliability by introducing a new commit orchestration pipeline. The new pipeline focuses on the instance document write path, and this PR both fixes the race, and integrates the new commit pipeline for rebase and push; fixing the optimization race.

The new `meta_commit_queue` module provides a recursive, per-database lock. The lock serializes writes to the shared `_meta` graph, system graphs, and any database-related graph, so that background optimization cannot change a graph head while a branch/database state-changing operation is in its commit window.

The following modules were changed:

1. **`src/core/document/meta_commit_queue.pl`** (new) – recursive, per-database lock (`with_meta_commit_lock/2`, `with_meta_commit_locks/2`), global mutex creation guard, and `graph_label_to_lock_key/2` to map any database-related graph label (including repository `_commits` and branch instance/schema graphs) back to the database lock key, while system graph labels map to a dedicated `system_meta` key.

2. **`src/core/transaction/database.pl`** – `run_transactions/4` acquires the lock for every database descriptor and system descriptor in the transaction set before `commit_validation_objects/2` runs. The lock is released before `post_commit_hook` so plugins such as the auto-optimizer do not deadlock.

3. **`src/core/api/db_rebase.pl`** – `rebase_on_branch/9` holds the database lock from the repository head update through the database commit. This is the exact window in which the optimizer was racing with the rebase.

4. **`src/core/api/db_push.pl`** – both remote URL and local shared-store push branches hold the same database lock across the repository-head update and the database commit.

5. **`src/core/api/api_optimize.pl`** – `named_graph_optimize/1` uses `graph_label_to_lock_key/2` to acquire the correct lock while reading the head, squashing, and setting the new head.

6. **`src/core/api/db_delete.pl`** – `delete_database_label/2` holds the database lock while calling `safe_delete_named_graph/2`, preventing the optimizer from touching a graph that is being deleted.

7. **`src/core/document/commit_queue.pl`** – new optimization scheduling state and active-branch tracking. Pending branch and database optimizations are recorded under the scheduler mutex; active-branch counts are incremented when `pick_next_branch/2` hands a branch to a worker and decremented in `requeue_branch/1` when the worker leaves. When a branch queue drains, any pending branch optimization is run under the mapped database lock; when the database active-branch count reaches zero, any pending database `_meta` optimization is run. Database-level optimizations for database keys that have no tracked branch activity (e.g. `system_meta`) are placed on a new global task queue and processed by idle commit workers. A deadline rule uses `*_due/1` flags so that a second optimization request for the same branch/database is scheduled for immediate execution in the commit flow as soon as the relevant queue drains, rather than being dropped or running synchronously in the caller thread.

8. **`docker/plugins/auto-optimize.pl`** – the probabilistic `maybe_optimize/1` trigger is kept, but when it fires it now calls `commit_queue:schedule_database_optimization/1` and `commit_queue:schedule_branch_optimization/1` for the descriptors in the validation objects instead of spawning a dedicated optimization thread. The `optimization_running` flag and `optimization_thread_wrapper` are removed; the commit queue enforces single-flight behavior and the deadline path.

## Commit queue alignment

- **Rebase now uses a verification contract.** In `src/core/api/db_rebase.pl`, the request thread does the expensive history computation and commit replay outside the lock, then submits a `rebase_contract` dict to the branch commit queue. The worker, which already holds the branch lock, verifies the target branch head, performs the final relink, and runs the database commit under the `meta_commit_lock`. If the head moved, the worker rejects the contract and the request thread immediately retries synchronously.
- **Explicit optimize is queued and deferred until idle.** The server route now calls `api_optimize_queued/3` in `src/core/api/api_optimize.pl`. For branch descriptors, this enqueues an `optimize_when_idle` contract. The worker only runs the rollup when the branch queue is empty; otherwise it puts the package back and continues processing commits. The worker retries immediately once if the label version changed, and returns the underlying `label_version_changed` error if the second attempt also fails, so the existing optimize error handler formats it correctly.
- **Commit-queue dispatch is multi-type.** `src/core/document/commit_queue.pl` no longer assumes every queued package is a document commit. It inspects `package_type` and dispatches to `run_commit_package`, `run_rebase_contract`, or the idle-defer path for `optimize_when_idle`.
- **Synchronous fallback for tests.** When the worker pool is not running (e.g. during most PLUnit tests), rebase falls back to running the final commit window synchronously in the request thread, still under the `meta_commit_lock`. This preserves the existing unit-test behavior without requiring a worker pool.
- **Rebase error formatting is registered.** `src/core/api/api_error.pl` now declares `error_type_(rebase, 'api:RebaseErrorResponse')` so generic errors such as `commit_queue_timeout` that escape the rebase handler are serialized correctly.
- **Shared reply delivery.** `api_document:deliver_commit_result/2` is now exported and reused by the rebase and optimize worker predicates, removing the duplicate `deliver_rebase_result` and `deliver_optimize_result` predicates.
- **Fix error double-wrapping in the commit queue request path.** `api_document:handle_commit_result/3` was re-wrapping exceptions that were already `error(Formal, Context)` terms, causing unhandled 500 responses. It now throws the original exception term directly when it is already an error.

## Code review fixes

- **`src/core/document/commit_queue.pl`**
  - `branch_key_database_key/2` now parses the organization/database prefix of a resolved branch key and calls `organization_database_name/3`, so the active-branch count key matches the lock key used by `database_key_from_descriptor/2` and `meta_commit_queue`.
  - `cleanup_active_branches_for_worker/1` now decrements `database_active_branch_count` for every stranded `active_branch` entry it removes, preventing leaked counts when a worker is killed or crashes.
  - `schedule_database_optimization/1` now accepts repository descriptors (they are database-level work) and schedules them on a new global task queue for database keys that have no tracked branch activity, such as `system_meta`. Idle commit workers process the global queue instead of running the optimization synchronously in the scheduling thread.
  - `schedule_branch_optimization/1` is now restricted to branch descriptors only; repository descriptors are no longer misplaced on a branch queue that will never drain.
  - Added tests for the global task queue path and the due-flag deadline behavior, and updated the active-branch count and database-key tests to use the lock-key format.

The lock is recursive so that a high-level operation like rebase or push can hold the lock while `run_transactions/3` re-enters it for the database transaction object.

## Regression tests

- `rebase_races_with_database_optimize` in `src/core/api/db_rebase.pl` reproduced the original race deterministically. It runs the rebase while a background thread optimizes the `admin/foo/_meta` descriptor. Before the fix, the rebase fails; after the fix, the rebase completes successfully and the optimizer thread exits cleanly. It is placed in a separate `rebase_race` test block with `[concurrent(false)]` because it uses a global hook and thread registration that must not leak into other concurrently running rebase tests.

- New `meta_commit_queue` unit tests in `src/core/document/meta_commit_queue.pl` cover system graph label mapping, database/branch/repo graph label mapping, and recursive same-thread locking.

- New `commit_queue` scheduling tests in `src/core/document/commit_queue.pl` cover database-key derivation, active-branch count tracking and the negative-count invariant, pending/due scheduling state, branch optimization running when the branch queue empties, the deadline rule for a second branch optimization, database optimization running only when the database has no active branches, and the global task queue for system/untracked database optimizations.


hoijnet:
This is resolved and has new tests, as part of the release candidate build for 12.1

PR #2463, opened 2026-07-04
Upgrade patch/diff to preserve high precision, and use the Rust JSON parser for performance

### LINKED ISSUE #2442: Patch fails for decimal properties
Trying to patch a decimal property fails.

**To Reproduce**
1. Create a test schema class:
```
{
     "@id": "Test",
     "@type": "Class",
     "weight": "xsd:decimal"
}
```
2. Create a class instance
```
{
     "@type": "Test",
     "weight": 1.1
}
```
3. Try to patch the weight property using the HTTP "patch" endpoint
```
{
    "patch": [
        {
            "@id": "Test/ORsQGF7ECTy8va5w",
            "weight": {
                "@after": 1.2,
                "@before": 1.1,
                "@op": "SwapValue"
            }
        }
    ],
    "author": "foo",
    "message": "Patch weight."
}
```
4. The API returns a conflict error although the expected and found value are identical.
```
{
    "@type": "api:PatchResponse",
    "api:error": {
        "@type": "api:PatchConflict",
        "api:conflicts": [
            {
                "@id": "Test/ORsQGF7ECTy8va5w",
                "weight": {
                    "@expected": 1.1,
                    "@found": 1.1,
                    "@op": "Conflict"
                }
            }
        ]
    },
    "api:message": "The patch did not apply cleanly because of the attached conflicts",
    "api:request_id": "761c7fae-5484-11f1-8ba6-0242ac120004",
    "api:status": "api:conflict"
}
```

**Info**
OS: Windows 10
How did you run terminus-server: Docker
TerminusDB Version: 12.0.4

hoijnet:
TerminusDB uses arbitrary precision decimals internally and in all other JSON interfaces, but does not offer this native support in JSON diff and patch interfaces (yet). Another problem with the current JSON processing is that large diffs and patches are not possible.

The JSON support follows the ISO JSON standard. It is silent on how to represent decimal numbers, and our interpretation is that it should support the maximum allowed by the processing system. We process numbers as rationals by default and therefore use high precision decimals and integers on the wire when used internally, and lower precision when the data model allows for it. For lesser precision, use float or double.

* [ISO/IEC 21778:2017 - The JSON data interchange syntax](https://www.iso.org/standard/71616.html)

## Upgrade to high precision decimals

The patch endpoint was using the default Prolog JSON parser, which converts wire-level decimal numbers into IEEE-754 floats. When a stored `xsd:decimal` value was held as a rational, applying a patch that contained the same value as a float produced a `Conflict` because the two representations were not exactly equal (this was fixed in #2462). The diff endpoint had the same problem: it could only see the float-rounded view of submitted decimals.

However, there was ambiguity in the interfaces as they did not follow the canonical TerminusDB standards-compliant JSON form for precise decimals.

This change routes both `/api/patch` and `/api/diff` payloads through the Rust `json_preserve` parser, which keeps JSON numbers as Prolog rationals. Because the stored values are also rationals, `values_equal/2` can now compare them exactly and `SwapValue` operations apply cleanly.

## What changed

- `src/server/routes.pl`: added a Rust-parser payload reader (`http_read_json_preserve_semidet`) and wired both patch and diff handlers through it via the new `add_payload(json_preserve)` option in `cors_handler`.
- `src/core/document/patch.pl`: added a `simple_patch` unit test that exercises `json_preserve:json_read_dict/3` to prove that wire decimals round-trip into rationals.
- `tests/test/diff.js`: added a raw decimal `SwapValue` patch test.
- `tests/test/decimal-precision.js`: added a full lifecycle integration test that inserts a document with 20-digit decimals, computes a diff against a target document, and applies the patch. The test uses the raw diff response text when sending the patch body, so JavaScript float parsing does not truncate the high-precision values before they reach the server.


hoijnet:
Closing by including in 12.1-rc instead

PR #2486, opened 2026-07-20
Apply explicit diffs

Kidnix101:
Hi,
I tried to extent the functionality of the apply API endpoint. 

# Current Situation
At the moment, diffs can only be applied implicitly by specifying a before and after commit.

# Changes
I added an optional "diffs" parameter to the apply API endpoint. The new parameter takes priority over the specification of a before and after commit, but has no effect on current implementations of the API that do not use this parameter.

# Motivation
To my understanding, `patch/{path}` supports only patches to existing documents. The operations `Insert` and `Delete` are not supported.

`apply/{path}` supports these operations, but only implicitly via the before and after commit parameters.

# Use Case
We have a main and feature branch which share the following baseline commit
## Baseline
```
{
    "@id": "Person/A"
    "name": "Anna"
}
```
## Main Branch
Let's say we performed the following changes on the main branch:

- Update document `Person/A` with name "Monica"

## Feature Branch
Let's say we performed the following changes on the feature branch:

- Insert new document `Person/B`
- Update document `Person/A` with name "Laura"

## Merge Attempts
We want to merge the feature into the main branch. Due to conflicting changes in `Person/A` we have to find a solution to resolve the merge conflict.

#### Keep feature branch version
- [x] If we want to keep the version of `Person/A` ("Laura") from the feature branch, we can explicitly specify the `SwapValue` operation with "Monica" as expected before value:

```
{
    "@id": "Person/A",
    "name": {
        "@op": "SwapValue",
        "@before": "Monica",
        "@after": "Laura"
}
```
#### Keep main branch version
- [x] If we want to keep the version of `Person/A` ("Monica") from the main branch, we can simply omit the `SwapValue` diff

This works great to resolve our merge conflict.
#### Issue
**However, the insertion of `Person/B` fails, because the operation `Insert` is not supported**

#### Keep feature branch version
- [x] If we want to keep the version of `Person/A` ("Laura") from the feature branch, we call the `apply/{path}` endpoint with the feature branch head as `afterCommit`, the main branch head as `beforeCommit`and point the `path` to the main branch.
#### Keep main branch version
If we want to keep the version of `Person/A` ("Monica") from the main branch, we call the `apply/{path}` endpoint with the feature branch head as `afterCommit`, the baseline commit as `beforeCommit`and point the `path` to the main branch.
#### Issue
**Trying to keep the main branch version results in a merge conflict we can only resolve by creating a new commit on the feature branch**

### Solution: Apply with explicit diffs
We'll proceed as we did with the `patch` endpoint. However, since the `apply` endpoint supports `Insert` and `Delete` operations, we run into no issues.

As this is my first PR for TerminusDB, please feel free to offer suggestions or criticism.
Thanks in advance!


Kidnix101:
@hoijnet I tried to take inspiration from your PR #2464 . Would be great if you could have a look and provide feedback. Thanks! :)

hoijnet:
Thanks for the PR, looking forward to reviewing in detail. I'm trying to understand what additional benefit with the apply diffs over the existing `/patch/{path}`, that sounds similar?

> One important use case I see for this change is that it makes it easier to handle merge conflicts.

Is the reason an ability to apply the patches individually or something else? I have not looked that deep yet into your new code but it looks similar to what patch offers, also it will be good to document together with an example.

Please include (in the PR text) an overview or example of the advantages with this construction over the existing `/patch/{path}` if you can?

My first gut instince is that patch should take patch objects, and apply takes a sequence of already committed objects and apply them elsewhere, but I will need to read up a bit more on the docs too!

Kidnix101:
@hoijnet I updated the PR description. Could you have a look?

hoijnet:
I'm mulling over it wherher this is apply or if we should be extending diff and patch with Insert and Delete operations which intuitively (to me) feels like it makes more sense?

Apply is in my mind to apply changes sourced from existing commits; and patch to introduce changes to a branch.

Not sure yet. Would it make sense to you if we extended patch with insert/delete instead of opening up for the operations on apply?

I'll take a deeper look too!

Kidnix101:
That's definitely a valid option as well. I opted for the `apply`route approach because it didn't require any changes to the internal logic; it simply allows the end user to use the internal function `apply_diff` directly.
`patch`, on the other hand, seems (at least in its current implementation) to be intended only for changes to existing resources, especially given that patch can also be applied to hypothetical documents and does not necessarily have to point to a commit or branch.

hoijnet:
Pros and cons have now been considered, I lean towards patch rather than apply. The reason is that diff today produces a patch that patch can't accept if it contains Insert or Delete; which does not really make sense.

Going with Apply would work, but it would be unclear to the user. Therefore I believe the best option is to keep the expected symmetry between diff and patch; and bring full support for patching what comes out of diff. Means a little more work before this PR is done, but I hope to make a contribution soon!

hoijnet:
@Kidnix101 , submitted a PR on your PR with the suggestion to move the functionality to the patch endpoint. There is a separate PR for build verification of it that I will close once the PR is accepted into your branch.

https://github.com/Kidnix101/terminusdb/pull/1

Would it work for your needs? This is great improvement as it brings parity to apply any diff, as insert and delete were not supported in the past.

Kidnix101:
@hoijnet  Thank you so much! I closed your PR.

hoijnet:
> @hoijnet Thank you so much! I closed your PR.

Thanks! Did you get a chance to test it on your end yet so it fulfills your needs and works as expected?

hoijnet (approved):
Approved. Thanks for the PR!

hoijnet:
@Kidnix101 the patch fix is now part of the latest `terminusdb/terminusdb-server:dev` dev container image.

* https://hub.docker.com/layers/terminusdb/terminusdb-server/dev/images/sha256-f4c12fb9c20d977813d5ea87724aa346ef11b715b568ffb243684b74ac5b1e05

Please share how it goes!

Kidnix101:
@hoijnet I tested the dev image. Works like a charm! :)

hoijnet:
Thanks for the contribution!! The latest 12.0.7 release is now out since today and the upcoming 12.1 should go RC relatively soon!

PR #2501, opened 2026-08-05
GraphQL SSE cleanup round

hoijnet:
Clean up and align with standards.

Also switch from descriptor-based limits to per-user subscription limits.


hoijnet:
This PR should have been against 12.1-rc instead of main, archiving and creating a new one against 12.1-rc.

PR #2505, opened 2026-08-05
Upgrade to swipl 10.0.2 due to addressing updateAlerted

hoijnet:
The 12.1-rc version needed the update to swipl 10.0.2 and this is a good time to bring the latest version into the next patch release too.


hoijnet:
Closed again, needed a different approach. PR coming.

